PR #9, opened 2013-01-15
Add two new module

takaki:
I wrote two module; Finance::Quote::Infobank and Finance::Quote::Kdb.
These are for Japanese mutual fund and stock price.


vincentl:
This merge request is several years old and conflicts with the current master.  Please open a two new merge requests (one for Finance::Quote::Infobank and one for Finance::Quote::Kdb) if those modules are still in use.

PR #27, opened 2015-02-03
Fidelity fixed

mta1309:
This series of changes addes the module originally written by Peter Ratzlaff to get bond quotes from Fidelity.  I've been using it for several years and have updated it over time to keep up with changes on the Fidelity web site.  If you want I can collapse all these into one commit, or you can leave them as is to preserve the history.

The module name (fidelity_fixed) may not be the best, but that's the name it had when I got it.  Change it if you want to.

### DIFF
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -174,7 +174,7 @@ sub new {
     shift(@reqmodules) if (@reqmodules);
     # Default modules
     @modules = qw/AEX AIAHK ASEGR ASX BMONesbittBurns BSERO Bourso
-            Cdnfundlibrary Citywire CSE Currencies Deka DWS FTPortfolios Fidelity
+            Cdnfundlibrary Citywire CSE Currencies Deka DWS FTPortfolios Fidelity FidelityFixed
             FinanceCanada Fool FTfunds HU GoldMoney HEX IndiaMutual LeRevenu
             ManInvestments Morningstar MorningstarJP MStaruk MtGox NZX Platinum
             SEB SIXfunds SIXshares StockHouseCanada TSP TSX Tdefunds
--- a/lib/Finance/Quote/FidelityFixed.pm
+++ b/lib/Finance/Quote/FidelityFixed.pm
@@ -1,0 +1,181 @@
+#!/usr/bin/perl -w
+#
+# FidelityFixed.pm
+#
+
+# Version 1.01
+# Modification of Rolf Endres' Finance::Quote::ZA
+#
+# Peter Ratzlaff <pratzlaff@gmail.com>
+# 2012.01.04
+
+package Finance::Quote::FidelityFixed;
+require 5.004;
+
+use strict;
+use vars qw /$VERSION/ ;
+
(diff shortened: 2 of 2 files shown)


ecocode:
hi,

I tried this module with the fidelity.t test. However almost all tests failed. Maybe the symbols are wrong ? That would be an easy fix :). Can you test on your system and maybe send me other symbols which _do_ work now ?

mta1309:
--On February 18, 2015 at 12:48:19 PM -0800 Erik Colson 
<notifications@github.com> wrote:

> I tried this module with the fidelity.t test. However almost all
> tests failed. Maybe the symbols are wrong ? That would be an easy fix
> :). Can you test on your system and maybe send me other symbols which
> _do_ work now ?

I think you misunderstand the purpose of that module, which may be 
badly named.  It isn't a replacement for the Fidelity module and 
doesn't use the same symbols.  Instead it is a new module that 
retrieves prices for municipal bonds and similar investments.  I didn't 
pick the name so I don't know the rationale for it, but it was probably 
chosen because these sorts of things pay a fixed return.  The page at 
<https://www.fidelity.com/fixed-income-bonds/overview> is Fidelity's 
main page for this sort of investment.

The symbols used by this module are the CUSIP codes for the securities 
in question.  Some that work are 59333PRJ9, 971175PC2, and 594712RK9. 
These are for "Miami Airport Bond 5.25% 10/1/20", "Willow Run 4.125% 
5/1/27", and "MSU Series C 4.5% 2/15/35" respectively.

The original code from Peter Ratzlaff didn't include a test suite and I 
haven't written one.  Part of the reason is that I've never gotten F::Q 
quote to build from the GIT source.  I'm not an expert on Perl and 
don't understand the new build system you switched to a while back.  I 
tried briefly to get it going but it did't seem to play well with 
MacPorts which is what I use to install Perl and the various modules I 
use.  I could abandon this and install Perl separately, but haven't 
done so since it's more convenient to keep everything in one dependency 
system.  If I installed Perl separately I would likely end up with two 
copies of many things, one installed by MacPorts as a dependency of 
some non-perl program and one installed via CPAN or whatever.

This is not so relevant now since I recently discovered that you had 
hooked F::Q up to the Travis CI system and I created an account on it 
to use to run the F::Q tests.  Now that I've gotten that working, I 
should create a test file for fidelityfixed.  I was more interested in 
seeing if the TIAA/CREF tests would work with my replacement module 
based on Yahoo.  They didn't work initially which is part of the reason 
that there is a second commit in that pull request.

               Mike

ecocode:
Hi Mike,

I modified the source a bit and merged into master. Also added a testfile.
Can you check it out ?

mta1309:
--On February 19, 2015 at 1:24:35 AM -0800 Erik Colson 
<notifications@github.com> wrote:

> I modified the source a bit and merged into master. Also added a
> testfile. Can you check it out ?

I gave it a try and it seems to work fine.  Thanks for writing the test 
file.

           Mike

ecocode:
OK.
Thanks for submitting/testing the module !

PR #31, opened 2015-03-18
Hard coding currency for YahooJSON because yahoo doesnt include it in the JSON

blm:
I added a hash that maps the Stock Exchange suffix to a currency.  If the stock has a suffix in the hash, an extra currency field is passed back to the calling function.

This is all to work with GNUCash.  I dont have GNUCash so I havent tested it directly.

This patch is supposed to fix [102637](https://rt.cpan.org/Ticket/Display.html?id=102637)

Comments and suggestions to improve are welcome.


ecocode:
thanks!
Support is added in version 1.38 in a slightly different way based on your code.

PR #32, opened 2015-04-05
Tiaacref fixes

mta1309:
The data returned from the web site for the old fake symbols used by the TIAA/CREF module changed.  The first field is now the new symbol for the fund in question instead of the name of the fund.

Also move the test of $check{$line[0]} inside the if that makes sure that it is defined.  This avoids an error if the web site returns a value for a security we didn't ask for.


ecocode:
merged !

PR #35, opened 2015-08-24
removing indirect object notation

garu:
Hello!

Thanks for taking the time to work on Finance::Quote! This is a very simple patch to replace any uses of indirect object notation:

    my $obj = new Some::Class

with the recommended Perl notation:

    my $obj = Some::Class->new

Hope it helps! Thanks!


ecocode:
merged !

garu:
Wow, that was fast! Thanks!

PR #36, opened 2015-12-24
Update docs, add prereq & strict

bpa:
Update doc and README to reflect change to use Dist::Zilla
Add strict on two modules to pass Kwalitee test
Add test prerequisite to dist.ini

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,10 @@ the empty list may be returned, or undef in a scalar context.
 
 # INSTALLATION
 
-To install this module, run the following commands:
+To install this module, make sure you have Dist::Zilla installed, and then run the following commands:
 
-    perl Makefile.PL
-    make
-    make test
-    make install
+    dzil test
+    dzil install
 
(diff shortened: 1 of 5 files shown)


ecocode:
thanks!

Can you specify any reason why you added :

  [Prereqs / TestRequires]
  HTML::TableExtract = 0

Since we already added AutoPrereqs this dependency should be auto detected..

bpa:
I think I must have using dzil wrong at first and thought it wasn't picking
it up.  I'm seeing HTML::TableExtract in `dzil listdeps` without that
change.

I also forgot to mention this was part of the 2015 Pull Request Challenge.
I hadn't used Dist::Zilla before checking this out.  No Makefile.PL threw
me at first, but cpan and cpanm have no problems building for me.

vincentl:
This pull request is several years old and the Documentation has changed since then.  Please open a new pull request if the current documentation is insufficient.

PR #38, opened 2016-02-01
Update TSX Test Stocks

rdswift:
Updated the securities used to test the tsx.pm module.  Some of the original symbols listed are no longer active for the securities listed in the test.  These would always show as failing the test.


ecocode:
Merged !

Thank you

PR #40, opened 2016-08-09
Add support for Bloomberg

alex314159:
This will allow download of prices from the www.bloomberg.com website.

### DIFF
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -173,7 +173,7 @@ sub new {
   if (!@reqmodules or $reqmodules[0] eq "-defaults") {
     shift(@reqmodules) if (@reqmodules);
     # Default modules
-    @modules = qw/AEX AIAHK ASEGR ASX BMONesbittBurns BSERO Bourso
+    @modules = qw/AEX AIAHK ASEGR ASX BMONesbittBurns Bloomberg BSERO Bourso
             Cdnfundlibrary Citywire CSE Currencies Deka DWS FTPortfolios Fidelity FidelityFixed
             FinanceCanada Fool FTfunds HU GoldMoney HEX IndiaMutual LeRevenu
             ManInvestments Morningstar MorningstarJP MStaruk NZX Platinum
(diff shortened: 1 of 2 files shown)


bpschuck:
It appears the Bloomberg API has since changed. While attempting `gnc-fq-dump bloomberg RGAGX:US` it returns `Can't call method "as_text" on an undefined value at /Users/bruceschuck/PerlLibs/lib/Finance/Quote/Bloomberg.pm line 47.`

Dumping the output of what the URL returns:
`
<!doctype html>
<html>
<head>
    <title>Bloomberg - Are you a robot?</title>
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="stylesheet" type="text/css" href="https://assets.bwbx.io/font-service/css/BWHaasGrotesk-55Roman-Web,BWHaasGrotesk-75Bold-Web,BW%20Haas%20Text%20Mono%20A-55%20Roman/font-face.css">
.
.
.
`

alex314159:
It appears Bloomberg have identified our id as not complying with their policy (they don’t want bots). I need to find a better way to circumvent the captcha. At this point all I can suggest is you try from a mobile hotpoint (your phone for example) that sometimes does the trick

bpschuck:
I was only trying it to check to see if it should be merged into F::Q's master branch while reviewing some old pull requests for clean up. Successful data retrieval should be consistent regardless of who uses it and where. In addition, there is no respective test script in your branch.

alex314159:
Bloomberg have indeed changed their web page. This code:
[https://github.com/alex314159/PerlFinanceQuoteBloomberg]
until three weeks or so ago when Bloomberg caught me with their captcha again. Historically the fix has been to change the headers in the request, but would welcome help if someone can think of a more permanent fix.

bpschuck:
You would still need to create a test file t/bloomberg.t. 
While not a hard requirement, you should extract the last trade date from the data received. Otherwise the date the data was retrieved is used by F:Q/GnuCash as the pricing date.
Thank you.

PR #41, opened 2016-08-21
Add missing column '% Chg' (p_change) to ASX.pm. Bug 97551.

goodvibes2:
p_change now comes from ASX instead of being calculated.
This missing column was causing following fields values to be in
wrong fields.


goodvibes2:
As from ~ 21 Apr 2017 ASX.pm needed a new URL (see Issue 46 https://github.com/finance-quote/finance-quote/issues/46), so this pull request (#41) is cancelled.
The fix for Issue 46 is pull request 47 https://github.com/finance-quote/finance-quote/pull/47 which also fixes bug 97551 https://rt.cpan.org/Public/Bug/Display.html?id=97551

Issue #42, opened 2016-08-22
New test in asx.t getting Use of uninitialized value $got in string eq at (eval in cmp_ok)


goodvibes2:
Hi,

I've modified ASX.pm to add currency AUD for some specific asx codes which are not indexes and it works fine but when I put a new test in asx.t it fails with:
cgood@goodi7vmutest:/home/cgood/github/finance-quote/t
$ ONLINE_TEST=1 ./asx.t
1..35
ok 1 - Data returned for call to asx
ok 2 - WES query was successful
ok 3 - Last price for WES was > 0
ok 4 - BHP query was successful
ok 5 - Last price for BHP was > 0
ok 6 - Data returned for call to fetch
ok 7 - BOQ query was successful
ok 8 - Last price for BOQ was > 0
ok 9 - Currency of BOQ is AUD
Use of uninitialized value $got in string eq at (eval in cmp_ok) ./asx.t line 46.
not ok 10 - Currency of XRO is AUD
#   Failed test 'Currency of XRO is AUD'
#   at ./asx.t line 46.
#          got: undef
#     expected: 'AUD'
ok 11 - No percentage sign in p_change value
... etc (other tests all pass)

I cannot understand what is wrong.

Here is the new test in asx.t:
# Check that we're getting currency information.
cmp_ok( $quotes{"BOQ", "currency"}, "eq", "AUD"
      , "Currency of BOQ is AUD" );

# Check that we're getting currency information for share starting with X.
cmp_ok( $quotes{"XRO", "currency"}, "eq", "AUD"
      , "Currency of XRO is AUD" );

The first test for BOQ is the original test which works fine. I just copied those lines and changed BOQ to XRO. XRO works fine when I use GnuCash: gnc-fq-dump asx XRO (apart from usual no date being returned by ASX which is taken care of by GnuCash - No quotes from asx return a date).

I'd really appreciate any help!

Finance::Quote 1.38
Test::More CPAN_VERSION 1.302052 INST_VERSION 1.001014
Test::Builder CPAN_VERSION 1.302052 INST_VERSION 1.001014
Ubuntu 16.04

[asx.t](http://pastebin.com/wpYwwtbk)
[ASX.pm](http://pastebin.com/7QBycrw0)

Regards, Chris Good

PR #47, opened 2017-05-09
Issue 46 Use new m.asx.com.au url as old no longer works

goodvibes2:
Also add currency AUD to stocks starting with X which are not indexes. (GnuCash bug https://bugzilla.gnome.org/show_bug.cgi?id=770132 - Price editor does not find ASX:XRO)

As a side benefit, this also fixes https://rt.cpan.org/Public/Bug/Display.html?id=97551 : missing column '% Chg' (p_change).

### DIFF
--- a/lib/Finance/Quote/ASX.pm
+++ b/lib/Finance/Quote/ASX.pm
@@ -32,6 +32,7 @@
 require 5.005;
 
 use strict;
+use warnings;
 
 package Finance::Quote::ASX;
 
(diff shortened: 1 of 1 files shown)


ecocode:
Hello Chris,
I downloaded your pull-request and ran the asx.t test file on it.  It looks like only a few tests pass.  Can you check the test-file and eventually correct it if it contains erroneous symbols ?

best

goodvibes2:
Hi Erik,

Thanks very much for merging my pull request.
I have raised a github issue previously because the testing was failing when it should be passing and I could not figure out why.
I will review the symbols asap.

Regards,
Chris Good

goodvibes2:
Issue with test failing is https://github.com/finance-quote/finance-quote/issues/42
I've figured out the problem with Issue 42 and will be modifying PR 80 soon to include extra tests.
asx.t tests do not contain any erroneous symbols.

PR #50, opened 2017-07-12
YahooJSON.pm - Hardcode 'EUR' currency to several european stock market suffixes

geisch:
In continuation of #31 :

Added several 'EUR' as currency to several other european stock market locations to the existing ones, following the listed abbreviations here:
http://search.cpan.org/dist/Finance-Quote/lib/Finance/Quote/Yahoo/Europe.pm

Suffixes '.CO' (Copenhagen, Denmark), '.L' (London, UK), '.O' (Oslo, Norway) and '.ST' (Stockholm, Sweden) trade in different currencies for which I do not know the abbreviation

### DIFF
--- a/lib/Finance/Quote/YahooJSON.pm
+++ b/lib/Finance/Quote/YahooJSON.pm
@@ -41,10 +41,21 @@ my %suffix_to_currency = (
     NS => 'INR',
     CL => 'INR',
     BO => 'INR',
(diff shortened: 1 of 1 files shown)


ecocode:
YahooJSON gets currency from the yahoo api.  No need to hardcode this ourselves.

PR #51, opened 2017-11-04
[WIP] Initial pass at adding AlphaVantage module to Finance::Quote

atmattpatt:
Hey there, like many others I was frustrated that Yahoo shut down their stock price API this week, thus rendering `Finance::Quote::Yahoo` unusable.

In the `finance-quote-devel` mailing list, @ecocode mentioned starting work on a dedicated module for [AlphaVantage](https://www.alphavantage.co) when time became available to him. While this is far from a finished product, I hope it provides a good enough start towards getting such a module included in `Finance::Quote`.

A couple notes:
- All of the tests for AlphaVantage pass given that `ALPHAVANTAGE_API_KEY` is set in the environment.
- I have not yet tried wiring this module up to GnuCash.
- I do not know if `Finance::Quote` has any opinions on how to store configuration such as an API key. I chose to use the environment because it was easy for me, but I'm not intimately familiar with how well that works on other platforms (specifically Windows).
- I definitely don't write Perl regularly, so the code itself could probably use some cleaning up. Large parts of it were unapologetically borrowed from the `Finance::Quote::YahooJSON` model. I am happy to take suggestions for how to improve the code, or to let someone else take this across the finish line.

Cheers!
-mattpatt

### DIFF
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -173,7 +173,7 @@ sub new {
   if (!@reqmodules or $reqmodules[0] eq "-defaults") {
     shift(@reqmodules) if (@reqmodules);
     # Default modules
-    @modules = qw/AEX AIAHK ASEGR ASX BMONesbittBurns BSERO Bourso
+    @modules = qw/AEX AIAHK AlphaVantage ASEGR ASX BMONesbittBurns BSERO Bourso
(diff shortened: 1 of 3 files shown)


ecocode:
I added your module. Thanks for your work!
Some modifications where added.  Basically don't use INTRADAY but DAILY to make volume, last, open return expected values

PR #52, opened 2017-11-04
Adding a historical MorningstarUS module

byteit101:
Now that Yahoo support vanished this week, I was looking at the list and saw that people were working on AlphaVantage, but was dismayed that everyone would have had to get their own API key. (Also it seems that some popular vanguard funds [might not be supported](https://sourceforge.net/p/finance-quote/mailman/message/36103598/)?) Searching the existing providers I was surprised that Morningstar had some non-us variants, but no US variant. So I found that there is a[ similar unofficial US API](https://gist.github.com/hahnicity/45323026693cdde6a116) that was very easy to parse, so created this module.
Some notes:

- This doesn't require an API key
- I managed to get this working in GnuCash by selecting it under the unknown category
- Supports all the Vanguard and Fidelity mutual funds and ETFs I have plus about half a dozen less-popular funds
- This is the first time I've written Perl, so I'm sure there are some cleanups that could happen. I used the `Morningstar` (confusingly for morningstar.se, GnuCash labels it "Morningstar, SE") module as a starting point (maybe that could be renamed? not sure, I didn't want to touch it)

Any feedback is much appreciated,
Patrick

### DIFF
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -176,7 +176,7 @@ sub new {
     @modules = qw/AEX AIAHK ASEGR ASX BMONesbittBurns BSERO Bourso
             Cdnfundlibrary Citywire CSE Currencies Deka DWS FTPortfolios Fidelity FidelityFixed
             FinanceCanada Fool FTfunds HU GoldMoney HEX IndiaMutual LeRevenu
-            ManInvestments Morningstar MorningstarJP MStaruk NZX Platinum
+            ManInvestments Morningstar MorningstarUS MorningstarJP MStaruk NZX Platinum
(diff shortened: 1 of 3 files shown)


vincentl:
It has been quite a while since this PR was open and now morningstar has changed.  I'm not familiar with morningstar and not sure if the intent of this PR was to fetch from https://www.morningstar.com/ or if there is/was a specific US site.

There are several morningstar versions in FQ, some working, some not working.

If anyone following this PR could comment on the continued desire for MorningstarUS.pm, whether it is supposed to fetch from https://www.morningstar.com/, and how we might manage the different morningstar variants, that would be appreciated.

byteit101:
In the intervening 3 years this this api has changed and I've moved to a different provider, so I think this can be closed at this point

PR #55, opened 2017-11-07
Check for error return from server.

mtalexander:
This is a trivial change to catch an error from the  AlphaVantage server (which keeps returning 503  errors for me).


zinglo:
I noticed that when the server was getting over loaded, it also was returning a 200 with HTML only (no JSON data).   JSON::decode_json just craps out and scripts die with an error.  It might also be a good idea to verify JSON data as well.   It could be something as trivial as (oversimplifying for example here)
eval {
   $json_data = JSON::decode_json $body;
   1;
} or do {
   $info{ $stock, 'success' } = 0;
};

PR #57, opened 2017-11-07
Add currency for Toronto (Canada) market.

sean-wallace:
Caveat: I haven't tested the changes to the test file.


ecocode:
thanks for your commit !

PR #67, opened 2017-11-19
Added date/time of last update and general cleanup data

frankcrawford:
This addresses the issue that there is no date returned by the ASX module, even though the web page indicates either the date of the last close or the current date/time.  Note that the time is only approximate, as the ASX delay reports by around 20 mins.  As well, the end of day is put arbitrarily at 6pm.

It also cleans up some of the other returned data to be numeric, returns a company name, and lists all the labels available from this module.

### DIFF
--- a/lib/Finance/Quote/ASX.pm
+++ b/lib/Finance/Quote/ASX.pm
@@ -41,6 +41,7 @@ use LWP::UserAgent;
 use HTML::TableExtract;
 use Encode;
 use HTTP::Cookies;
+use HTML::Parser;
 
 use vars qw/$ASX_URL @ASX_SEC_CODES/;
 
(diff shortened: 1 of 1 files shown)


goodvibes2:
Unfortunately this pull request is based on an ASX.pm which uses m.asx.com.au which stopped working approx 11 Feb 2018. There is a new url that may work:  https://search.asx.com.au/s/search.html?query=CBA+BOQ&collection=asx-meta&profile=web - I will investigate shortly, after I have created a pull request to fix the perl warnings from yahooJSON.pm which currently causes GnuCash to hang in Windows when getting quotes for multiple (>15 in my case) stocks.

frankcrawford:
Yep, they seem to have broken it.  I'll also have a look at that URL when I get a chance.

goodvibes2:
As this PR is now out of date, I suggest it be cancelled.

ecocode:
Thanks Chris for your followup.  Sorry to Frank as I was not able to include your work when it was solving issues.

frankcrawford:
Thanks, understand, since the ASX broke things again.

PR #69, opened 2017-11-20
URL changed as old URL for NAV no more works

amishxda:
Old URL no more works.

This patch uses more official URL which is obtained directly from AMFI India website. So it is more likely not to fail and not to change in future.


eternaltyro:
This. I came here looking to fix this all the way from GnuCash IRC. Fingers Crossed.

Issue #78, opened 2018-02-28
yahooJSON.pm returns date 00/00/2000 in Windows


goodvibes2:
This is because Microsoft strftime function does not support %T.
Also, because of multiple 'Use of uninitialized value' perl warnings, this causes GnuCash in Windows to hang when getting multiple quotes (in my case >15).

echo (yahoo_json "CBA") | perl gnc-fq-helper
Use of uninitialized value $year in numeric lt (<) at C:/strawberry/perl/site/lib/Finance/Quote.pm line 757.
Use of uninitialized value $month in pattern match (m//) at C:/strawberry/perl/site/lib/Finance/Quote.pm line 768.
Use of uninitialized value $month in sprintf at C:/strawberry/perl/site/lib/Finance/Quote.pm line 773.
Use of uninitialized value $day in sprintf at C:/strawberry/perl/site/lib/Finance/Quote.pm line 773.
Use of uninitialized value $month in sprintf at C:/strawberry/perl/site/lib/Finance/Quote.pm line 774.
Use of uninitialized value $day in sprintf at C:/strawberry/perl/site/lib/Finance/Quote.pm line 774.
[parse] Invalid date 00/00/2000 12:00:00
(("CBA" (symbol . "CBA") (gnc:time-no-zone . "2017-11-21 19:05:10") (last . 7.65) (currency . "USD")))

I shall soon create a pull request for fix these problems.

goodvibes2:
PR to fix this Issue is https://github.com/finance-quote/finance-quote/pull/79

ecocode:
Thanks ! merged

PR #79, opened 2018-03-02
yahooJSON.pm returns date 00/00/2000 in Windows

goodvibes2:
1. MS Windows strftime() does not support %T so use equivalent %H:%M:%S instead.
2. Fix perl warnings which can cause GnuCash to hang when multiple quotes (>15) are requested.


goodvibes2:
This PR is the fix for Issue https://github.com/finance-quote/finance-quote/issues/78

PR #82, opened 2018-03-05
Issue 80 ASX.pm m.asx.com.au no longer works

goodvibes2:
Go back to www.asx.com.au/asx/markets/priceLookup.do which now works again.
Unfortunately the following fields are no longer available (not needed for GnuCash):
date, name, div_yield, eps, pe + year_range.
There is a date available but it is always today, so this script does not return it (GnuCash gnc-fq-helper will default it to today).


ecocode:
thanks !

Issue #83, opened 2018-03-18
AlphaVantage sometimes fails - need to retry several times if fails


goodvibes2:
See https://sourceforge.net/p/finance-quote/mailman/message/36242664
Soon I will be submitting a Pull Request to change AlphaVantage.pm so that if it fails, it retries each stock up to 5 times, with a 20 second delay between each try, as suggested in the above message link.

goodvibes2:
Fix for this Issue is PR https://github.com/finance-quote/finance-quote/pull/84

PR #84, opened 2018-03-25
Fq iss83 alpha vantage : Retry up to 4 times if fails with Information message

goodvibes2:
Fix for Issue https://github.com/finance-quote/finance-quote/issues/83
1. Retry up to 4 times if fails with Information message with a 20 second wait between tries
2. Add 1 second sleep after each API call to share server resources as recommended by Alpha Vantage


goodvibes2:
The changes in this PR have also been included in Mike Alexander's PR https://github.com/finance-quote/finance-quote/pull/85 in which he also adds the 1 second delay between requests for currency quotes.

PR #86, opened 2018-03-31
update parsing for changed layout on dwsfunds source page

x42x64:
Extracting the table from the [DWS funds source page](https://www.deami.de/dps/ff/prices.aspx) using depth and count does not work anymore (layout change). Therefore the source dwsfunds is broken.

The fix can be as follows:
- Extraction of the table using the table id "FundsFinder_ResultTable".
- Replacing the call of the [deprecated `table_state` method](http://search.cpan.org/dist/HTML-TableExtract/lib/HTML/TableExtract.pm#DEPRECATED_METHODS) by `first_table_found()`

### DIFF
--- a/lib/Finance/Quote/DWS.pm
+++ b/lib/Finance/Quote/DWS.pm
@@ -127,9 +127,9 @@ sub dwsfunds {
 	if ($response->is_success) {
 		$html_string =$response->content;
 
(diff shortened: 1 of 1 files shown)


vincentl:
Tests passed!  Thanks for submitting the PR and sorry it has taken so long to merge.

Issue #56, opened 2017-11-07
Cannot follow install instructions - there is no Makefile.PM?


jusmop:
I cloned the git repository (master) but there is no Makefile.PM.
Can someone point me to it?

jusmop:
OK thanks.

I installed the latest, but I still cannot get stock quotes from within
GnuCash. I changed the quote source from Yahoo to ASX for every stock but
GnuCash reports that all quote requests failed.

I have tested the Finance-Quote stand-alone using a simple perl script and
it works.

How do I check if GnuCash is calling the correct package?

Cheers,
Justin.

ecocode:
jusmop <notifications@github.com> writes:

> OK thanks.
>
> I installed the latest, but I still cannot get stock quotes from within
> GnuCash. I changed the quote source from Yahoo to ASX for every stock but
> GnuCash reports that all quote requests failed.

If you want to change to the latest module alphavantage, the correct
source should be alphavantage, not ASX...

>
> I have tested the Finance-Quote stand-alone using a simple perl script and
> it works.
>
> How do I check if GnuCash is calling the correct package?

Can't help with that as I don't use GnuCash.  Please ask this question
on the GnuCash user list.

best
--
erik colson

jusmop:
Hi Erik,

Yes you are right - I need to ask my GnuCash questions on the GnuCash forum.

But I'm still confused by the source. The issue I was trying to understand
was that when I use F::Q in a simple perl script (see below), it *works*. I
get the latest ASX.NAB stock price. So what is the default source that F::Q
is using?

Surely it can't be alphavantage as posts on the GnuCash forum suggest that
using alphavantage requires installing a key, which I have not done...

#!/usr/bin/perl
use Finance::Quote;
my $q = Finance::Quote->new();
my %data = $q->fetch('ASX', 'NAB');
print "NAB price is " . $data{'NAB', 'price'} . "\n";

If I understood what the source is that F::Q uses in this script, then
maybe I could configure GnuCash to use the same, and then I'll be golden...


Cheers,
Justin.

ecocode:
jusmop <notifications@github.com> writes:

> Hi Erik,
>
> Yes you are right - I need to ask my GnuCash questions on the GnuCash forum.
>
> But I'm still confused by the source. The issue I was trying to understand
> was that when I use F::Q in a simple perl script (see below), it *works*. I
> get the latest ASX.NAB stock price. So what is the default source that F::Q
> is using?
>
[...]

PR #87, opened 2018-04-22
Update README.md

shaform:
according to #56


shaform:
close due to inactivity.

Issue #68, opened 2017-11-19
IndiaMutual / amfiindia URL changed


amishxda:
India Mutual Fund URL does not appear to be working anymore. Gives 500 Internal Server Error.

However setting it to this URL works
https://www.amfiindia.com/spages/NAVAll.txt

Source page:
https://www.amfiindia.com/nav-history-download

Filename = Finance::Quote::IndiaMutual.pm
Line to alter to:
$AMFI_URL = ("https://www.amfiindia.com/spages/NAVAll.txt");

amishxda:
PR: https://github.com/finance-quote/finance-quote/pull/69

amishxda:
Ah! Finally.. thank you! :+1:

ecocode:
sorry for the delay...

PR #90, opened 2018-07-12
Update IndiaMutual URL, csv format

shastry:
The URL update makes indiamutual work again.
The new csv updated the position of date field.

Test file fixes (if it's still being used).


shastry:
Also obsoletes #69 #68.

Issue #77, opened 2018-01-19
AlphaVantage now supports batch quotes


atmattpatt:
On Dec 28, AlphaVantage announced support for fetching quotes in batch. The returned data isn't as rich as the daily time series data, but it is significantly faster.

Example batch quote return value:
```json
{
    "Meta Data": {
        "1. Information": "Batch Stock Market Quotes",
        "2. Notes": "IEX Real-Time Price provided for free by IEX (https://iextrading.com/developer/).",
        "3. Time Zone": "US/Eastern"
    },
    "Stock Quotes": [
        {
            "1. symbol": "PPG",
            "2. price": "118.4100",
            "3. volume": "3734030",
            "4. timestamp": "2018-01-18 16:00:00"
        }
    ]
}
```

I think the main difference is that you only get a single `price` instead of `open`/`close`/`high`/`low`.

The speed increase is pretty significant though. Some benchmarking I did tonight shows it may be 29.6x faster for 25 symbols, 96.3x faster for 250 symbols. This is just from one run, but anecdotally the results were pretty consistent across multiple runs.

```
./benchmark.rb
Using 25 symbols: [list omitted for brevity]

Rehearsal ------------------------------------------------------------
Multiple single requests   0.090000   0.040000   0.130000 ( 29.845046)
Batch requests             0.000000   0.000000   0.000000 (  0.382820)
--------------------------------------------------- total: 0.130000sec

                               user     system      total        real
Multiple single requests   0.080000   0.040000   0.120000 ( 22.843711)
Batch requests             0.000000   0.010000   0.010000 (  0.772521)

Total number of requests executed (including rehearsal):
Single - 50
Batch  - 2

./benchmark.rb 250
Using 250 symbols: [list omitted for brevity]

Rehearsal ------------------------------------------------------------
Multiple single requests   0.890000   0.350000   1.240000 (237.607558)
Batch requests             0.010000   0.000000   0.010000 (  0.575303)
--------------------------------------------------- total: 1.250000sec

                               user     system      total        real
Multiple single requests   0.830000   0.320000   1.150000 (265.849517)
Batch requests             0.010000   0.000000   0.010000 (  2.759761)

Total number of requests executed (including rehearsal):
Single - 500
Batch  - 6
```
(Source is available at https://github.com/atmattpatt/alphavantage-benchmark)

[...]

PR #95, opened 2018-07-17
Add Alphavantage batch quote support

atmattpatt:
Per Alphavantage limitations, batch quotes only work for US securities and will not work for mutual funds. Batch quote support is added as a separate module, as these limitations may not make it suitable for all users (see https://github.com/finance-quote/finance-quote/issues/77#issuecomment-368998755)

I tried this out with Gnucash tonight and it does seem to be considerably more reliable when pulling quotes for all my securities. I'll continue to test it out over the coming days and weeks.

Fixes https://github.com/finance-quote/finance-quote/issues/77

### DIFF
--- a/lib/Finance/Quote/AlphaVantage_Batch.pm
+++ b/lib/Finance/Quote/AlphaVantage_Batch.pm
@@ -1,0 +1,195 @@
+#!/usr/bin/perl -w
+#    This module is based on the Finance::Quote::yahooJSON module
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


lib/Finance/Quote/AlphaVantage_Batch.pm
```diff
@@ -0,0 +1,195 @@
...
+                $info{ $stock, 'errormsg' } = "json_data doesn't contain Stock Quotes hash";
+            }
+            next;
+        }
+
+        foreach my $bq (@{$json_data->{'Stock Quotes'}}) {
+            foreach my $stock (@stocks_chunk) {
+                if ( $bq->{'1. symbol'} ne $stock ) {
+                    next;
+                }
```
atmattpatt: This has undesirably quadratic time complexity. However, my knowledge of Perl is still quite limited so this was the best that I could come up with. Edits/suggestions happily accepted 😄

atmattpatt:
It appears as AlphaVantage has removed this functionality

PR #103, opened 2018-10-31
Changed source for BVB exchange

Steinhagen:
Description:
The former link to obtain the info is no longer valid. Also, since the
bvb site no longer accesible using the perl python, I changed the
source to Tradeville.


vincentl:
Since this PR is a few years old, the site format changed.  Since the code needed updating anyways, I remained it to Tradeville.pm, to match the new source and removed BSERO.pm.

PR #108, opened 2019-03-16
feat: add support for retrieving quotes from morningstar.ch

fritteli:
I needed quotes for a fund that was only available on morningstar.ch, so I created this file, based on MStaruk.pm. I'd appreciate it if you included it, and I'm happy to respond to comments.


ecocode:
Hi fritteli,

Thanks for this work !

best

PR #115, opened 2019-06-15
Jsefix - Duplicate of Pull Request 109

bpschuck:
Pull request 109 was a few commits behind the master, including changes to the YahooJSON.pm module. In addition, there was a syntax error in the code added to that module in the pull request.
So it can be included in the next F::Q release, I created a new branch based on recent commits and added the code for correcting the currency returned for Johannesburg Stock Exchange (JSE) securities. It was easier to reconcile the merge using a current version of the module.

### DIFF
--- a/lib/Finance/Quote/YahooJSON.pm
+++ b/lib/Finance/Quote/YahooJSON.pm
@@ -148,6 +148,16 @@ sub yahoo_json {
                     $info{ $stocks, "currency"} = "GBP";
                 }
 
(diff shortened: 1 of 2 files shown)


bpschuck:
@ecocode - is it an issue that the CI fails for Perl 5.10 and 5.12. I'm not familiar with Travis CI/CD and not sure why the build fails for those two older versions of Perl. 
Thanks.

PR #130, opened 2019-08-06
Rest of fix for AlphaVantage rate limiting.

mtalexander:
Commit d392bbec missed a few places where Information needed to be
changed to Note.

### DIFF
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -274,9 +274,9 @@ sub currency {
     if ( !$json_data || $json_data->{'Error Message'} ) {
       return undef;
     }
-#     print "Failed: " . $json_data->{'Information'} . "\n" if (($try_cnt < 5) && ($json_data->{'Information'}));
-    sleep (20) if (($try_cnt < 5) && ($json_data->{'Information'}));
-  } while (($try_cnt < 5) && ($json_data->{'Information'}));
+#     print "Failed: " . $json_data->{'Note'} . "\n" if (($try_cnt < 5) && ($json_data->{'Note'}));
+    sleep (20) if (($try_cnt < 5) && ($json_data->{'Note'}));
+  } while (($try_cnt < 5) && ($json_data->{'Note'}));
 
   my $exchange_rate = $json_data->{'Realtime Currency Exchange Rate'}->{'5. Exchange Rate'};
 
--- a/lib/Finance/Quote/AlphaVantage.pm
+++ b/lib/Finance/Quote/AlphaVantage.pm
@@ -201,7 +201,7 @@ sub alphavantage {
 
         my $try_cnt = 0;
         while (($try_cnt < 5) && ($json_data->{'Note'})) {
-            # print STDERR "INFORMATION:".$json_data->{'Note'}."\n";
+            # print STDERR "NOTE:".$json_data->{'Note'}."\n";
             # print STDERR "ADDITIONAL SLEEPING HERE !";
             sleep (20);
             &$get_content();


bpschuck:
Rate limiting has already been addressed. In addition, the module has been changed to utilize a different API from AlphaVantage. Closing this pull request.

mtalexander:
This pull request is not irrelevant.  Except for a comment change which doesn't matter, it only affects the currency quote API which hasn't been changed and is still wrong.  At some point AlphaVantage changed the error message returned when you exceed their rate limit from a JSON "INFORMATION" field to a "NOTE" field.  This means that the currency quote API is not recognizing the rate limiting error message and is not limiting it's queries in response. to it.  The patch to lib/Finance/Quote.pm is still relevant and correct.  I'll submit a new pull request if that would help.

edward-dauvergne:
Please reconsider reopening this pull request and merging it!  The changes are essential.  I would even recommend releasing F::Q v1.50 with this and the previous fixes.  They allow, for example, GnuCash to actually fetch more than 5 currencies (the Alpha Vantage API limit when F::Q does not take the API limit into account).  I have created the following script that demonstrates the fix (assuming you have the ALPHAVANTAGE_API_KEY environmental variable set):

```
#!/usr/bin/env perl

use Finance::Quote;
use Time::HiRes qw(time);
print("Using Perl $^V\n");
print("Using Finance::Quote $Finance::Quote::VERSION\n\n");

# The currencies to fetch.
my @symbols = (
    "AUD",
    "BRL",
    "CAD",
    "CHF",
    "CNY",
    "CZK",
    "DKK",
    "EUR",
    "GBP",
    "HKD",
    "HRK",
    "ILS",
    "JPY",
    "KRW",
    "MYR",
    "NOK",
    "NZD",
    "RUB",
    "SGD",
    "TWD",
    "XAU"
);

# Fetch the currencies.
my $begin = time();
my $quoter = Finance::Quote->new();
foreach (@symbols) {
    $price = $quoter->currency("USD", $_);
    if (!defined $price) {
        $price = "Failed.";
    }
    print("USD$_: $price\n");
}
my $end = time();
printf("Fetching time %0.02f s\n", $end-$begin);
```

Using F::Q at bcfc947ee0173e059573236c2cf4a9bcb9a09b65, the result is:

```
Using Perl v5.30.1
Using Finance::Quote 1.50

USDAUD: 1.4909
USDBRL: 4.325
USDCAD: 1.3294
USDCHF: 0.9778
USDCNY: 6.9745
USDCZK: Failed.
USDDKK: Failed.
USDEUR: Failed.
USDGBP: Failed.
USDHKD: Failed.
USDHRK: Failed.
USDILS: Failed.
USDJPY: Failed.
USDKRW: Failed.
USDMYR: Failed.
USDNOK: Failed.
USDNZD: Failed.
USDRUB: Failed.
USDSGD: Failed.
USDTWD: Failed.
USDXAU: Failed.
Fetching time 12.44 s
```

After applying the changes in this pull request (locally via cherry-picking 877c87bd3ea1cbd403c59dfcd4a58b97c14d232a), the result is:

```
Using Perl v5.30.1
Using Finance::Quote 1.50

USDAUD: 1.4907
USDBRL: 4.3242
USDCAD: 1.3293
USDCHF: 0.9781
USDCNY: 6.9745
USDCZK: 22.833
USDDKK: 6.8465
USDEUR: 0.9162
USDGBP: 0.7734
USDHKD: 7.7646
USDHRK: 6.8309
USDILS: 3.4163
USDJPY: 109.85
USDKRW: 1183.9
USDMYR: 4.134
USDNOK: 9.2539
USDNZD: 1.5643
USDRUB: 63.644
USDSGD: 1.3872
USDTWD: 30.034
USDXAU: Failed.
Fetching time 219.69 s
```

vincentl:
Hi Edward - Just saw the thread on the GNC-dev mailing list about this issue.  I've been helping out on some FQ maintenance, but don't have any experience with the currency API.  We are discussing changes to the FQ API to address the need for API keys for various modules and a consistent naming scheme.  We didn't discuss the currency API, but it sounds like extending it to allow for multiple queries at once as well as catching the AlphaVantage currency error should be done.

I've reopened this pull request and will investigate over the next few days.

vincentl:
I'm guessing that the original design was for the use case where the fetcher is gathering various quotes in their native currency and all the quotes are converted to a specific target currency.  So you might have many "from" currencies, but a single "to" currency.

The example script that fails after 5 conversions is different because there are many "to" currencies.

Was the example simple for demonstration or is there a use case where there are multiple target currencies?

I am trying to sort out a few ideas.  Introducing CURRENCY_MODULES so that we can have different backends to fetch conversions instead of only using AlphaVantage.  How to restructure the main "fetch" routine so it can batch together all the conversion queries in one call. How to change or extend the "currency" function API.

bpschuck:
Erik also entered issue 126 last year to change the source to
https://www.iban.com/currency-codes. Not sure if he also meant to get
the exchange rates from there as well. Having it forced to be
AlphaVantange means that anyone that wants to have currencies converted
would need to be the API key. While it is free, some users may prefer
not to. So definitely being able to use other back ends is a good idea.

A minor point, but being a bit OCD I would think the same source should
be used in both Quote.pm and Quote/Currencies.pm.

Also note pull request 131 (related to issue 129), changes the
AlphaVantage URL used to fetch quotes. Instead of the TIME_SERIES_DAILY
api, it now uses the GLOBAL_QUOTE api.

- Bruce

mtalexander:
I don't think it matters whether all the from currencies are the same, all
the to currencies are the same, or they are random. It's just how many
currency requests you make in a short period of time that matters.

Improving the API would be good, but fixing the immediate problem shouldn't
wait for that.

The simplest change I've thought of to allow multiple exchange rates per
call is to allow one or both of the parameters to the currency function to
be a list. I'm not sure how hard that would be. Of course it wouldn't
matter unless there callers were changed too.

Mike

edward-dauvergne:
Here is the reverse script:

```
#!/usr/bin/env perl

use Finance::Quote;
use Time::HiRes qw(time);
print("Using Perl $^V\n");
print("Using Finance::Quote $Finance::Quote::VERSION\n\n");

# The currencies to fetch.
my @symbols = (
    "AUD",
    "BRL",
    "CAD",
    "CHF",
    "CNY",
    "CZK",
    "DKK",
    "EUR",
    "GBP",
    "HKD",
    "HRK",
    "ILS",
    "JPY",
    "KRW",
    "MYR",
    "NOK",
    "NZD",
    "RUB",
    "SGD",
    "TWD",
    "XAU"
);

# Fetch the currencies.
my $begin = time();
my $quoter = Finance::Quote->new();
foreach (@symbols) {
    $price = $quoter->currency($_, "USD");
    if (!defined $price) {
        $price = "Failed.";
    }
    print($_ . "USD: $price\n");
}
my $end = time();
printf("Fetching time %0.02f s\n", $end-$begin);
```

And its output:

```
Using Perl v5.30.1
Using Finance::Quote 1.50

AUDUSD: 0.6684
BRLUSD: 0.231
CADUSD: 0.755
CHFUSD: 1.0199
CNYUSD: 0.1428
CZKUSD: Failed.
DKKUSD: Failed.
EURUSD: Failed.
GBPUSD: Failed.
HKDUSD: Failed.
HRKUSD: Failed.
ILSUSD: Failed.
JPYUSD: Failed.
KRWUSD: Failed.
MYRUSD: Failed.
NOKUSD: Failed.
NZDUSD: Failed.
RUBUSD: Failed.
SGDUSD: Failed.
TWDUSD: Failed.
XAUUSD: Failed.
Fetching time 11.88 s
```

Mike's commit 877c87bd3ea1cbd403c59dfcd4a58b97c14d232a fixes both cases.

As I mentioned on the GnuCash mailing list (see the threads at https://lists.gnucash.org/pipermail/gnucash-devel/2020-January/044509.html and https://lists.gnucash.org/pipermail/gnucash-devel/2020-February/044522.html), if you have a close look at the `currency` subroutine in `Quote.pm` and the `alphavantage` subroutine in `Quote/AlphaVantage.pm`, you'll see that there is a large amount of duplicated logic for handling the API limits.  It would make sense to merge the Alpha Vantage code, as the `alphavantage` subroutine logic is more advanced, and shift it out of `Quote.pm` to make this module provider independent.  But this is a separate issue from this simple fix.

PR #137, opened 2020-02-03
New mutual funds source: Oslo stock exchange

sigvei:
This enables quotes from oslobors.no. The quotes are gathered from a
clean JSON source that is used internally by the (Angular-based)
website.

Provides a price, date, currency and p_change.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -457,6 +457,22 @@
   testfile: TBD
   testcases:
 #
+- module: Oslobors.pm
+  state: working
+  added: 2020-02-06
+  changed: 2020-02-06
+  removed:
+  urls:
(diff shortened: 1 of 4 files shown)


bpschuck:
Sigve,

Thank you for your contribution to Finance::Quote. Would it be possible to please add another 1 or two funds to the tests aside from OD-HORIA.OSE? Thanks.

Not a critical,s ince this is a new module, it would save the maintainers some time if you could also add a section to the Modules-README.yml file. Not a problem, we'll add it if you are unsure exactly what to add.

Thanks again, great work.

sigvei:
Okay, tests added and Modules-README updated.

bpschuck:
Failure in Travis-CI is because perlbrew returns an error installing Perl v5.28.

PR #138, opened 2020-02-19
Update Fidelity Direct source to make it functional

techdude:
- The old Fidelity link was broken, so I searched around and found the correct replacement which gives daily quotes for all Fidelity funds. 
- Some of the fields shifted around, so I updated the fields to match the new data source. One column (number) doesn't exist anymore, so I removed it (it's not needed by GnuCash).
- I updated the documentation to remove the note that this module isn't functional (since it is now functional).
- I tested that the module works with GnuCash 3.8b+ which I assume is the expected consumer of the information.
- The functionality can also be verified from a GNU cash install by running something similar to the following:
`perl gnc-fq-dump -v fidelity_direct FBGRX`

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -232,15 +232,20 @@
   testcases:
 #
 - module: Fidelity.pm
-  state: TBD
+  state: working
   added: TBD
-  changed: ~
+  changed: 2020-03-21
   removed: ~
   urls:
+    - https://fundresearch.fidelity.com/mutual-funds/fidelity-funds-daily-pricing-yields/download
   apikey: false
   notes: ~
-  testfile: TBD
+  testfile: fidelity.t
   testcases:
+    - FGRIX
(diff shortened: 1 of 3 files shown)


bpschuck:
@techdude - Thank you for your contribution to F::Q. I was looking over and testing the pull request. In the Fidelity.pm module, leave the `# VERSION` line as is please. The test file, t/fidelity.t also needs some changes. One change is removing the tests from the TODO block. The fund used to test getting the yield field returned needs to change. FSLXX is no longer a valid fund. I tested successfully with FEQTX. 
In Modules-README.yml, aside from removing the note about the module not working, adding the appropriate information for the state (working), changed (YYYY-MM-DD), urls, testfile, and testcases fields would be appreciated.

I can just as easily push these changes to your fork before merging the PR, but if you have time you can make these.

Thank you.

PR #150, opened 2020-08-06
Add suffixes FRK, DEX, AMS

morgil:
I encountered the suffix .FRK for Frankfurt, as well as DEX for Xetra and AMS for Amsterdam.

Example available at: https://www.alphavantage.co/query?function=SYMBOL_SEARCH&keywords=UIMM&apikey=[...]


vincentl:
Thanks!

PR #151, opened 2020-09-29
Fetching quotes from XETRA (Frankfurt stock exchange)

e-dorigatti:
Hi, I contributed a new module to fetch quotes from the [frankfurt stock exchange](https://www.xetra.com/xetra-en/), which provides a JSON endpoint. As I am new to perl, I based the module on `yahoo_json` and made the necessary modifications.

I don't know what is your policy on testing, but I can run the example script successfully:

```
$ perl stockdump.pl xetra DE000A0D8Q07
$VAR1 = {
          'DE000A0D8Q07exchange' => 'XETR',
          'DE000A0D8Q07high' => '35.42',
          'DE000A0D8Q07date' => '09/29/2020',
          'DE000A0D8Q07symbol' => 'DE000A0D8Q07',
          'DE000A0D8Q07isodate' => '2020-09-29',
          'DE000A0D8Q07method' => 'xetra',
          'DE000A0D8Q07close' => '35.445',
          'DE000A0D8Q07low' => '35.33',
          'DE000A0D8Q07last' => '35.42',
          'DE000A0D8Q07currency' => 'EUR',
          'DE000A0D8Q07success' => 1
        };

```

### DIFF
--- a/lib/Finance/Quote/XETRA.pm
+++ b/lib/Finance/Quote/XETRA.pm
@@ -66,7 +66,7 @@ sub xetra {
         else {
             $info{ $stocks, "success" } = 0;
             $info{ $stocks, "errormsg" } = "Invalid format: use only one '.' to separate ISIN from MIC";
-            continue;
+            next;
         }
 
         my $url = $PRICE_URL . "?mic=" . $mic . "&isin=" . $isin;
--- a/lib/Finance/Quote/XETRA.pm
+++ b/lib/Finance/Quote/XETRA.pm
@@ -1,0 +1,128 @@
+#!/usr/bin/perl -w
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program; if not, write to the Free Software
+#    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+#    02110-1301, USA
+
+package Finance::Quote::XETRA;
+
+require 5.005;
+
+use strict;
+use JSON qw( decode_json );
+use vars qw($VERSION $YIND_URL_HEAD $YIND_URL_TAIL);
+use LWP::UserAgent;
+use HTTP::Request::Common;
+use HTML::TableExtract;
+use Time::Piece;
+
+# VERSION
+
+
+my $PRICE_URL = "https://api.boerse-frankfurt.de/data/price_information?mic=XETR&isin=";
+
+sub methods {
+    return ( xetra => \&xetra,
+    );
+}
+{
+    my @labels = qw/exchange symbol last high low isodate date close currency/;
+
+    sub labels {
+        return ( xetra => \@labels, );
+    }
+}
+
+sub xetra {
+
+    my $quoter = shift;
+    my @stocks = @_;
+    my ( %info, $reply, $url, $te, $ts, $row, @cells, $ce );
+    my $ua = $quoter->user_agent();
+
+    # server reply is a perpetual event stream with continuous updates
+    # disconnect after we receive the first event
+    $ua->max_size(10);
+
+    foreach my $stocks (@stocks) {
+        my $url = $PRICE_URL . $stocks;
+        my $resp = $ua->request(GET $url);
+        my $code = $resp->code;
+        my $desc = HTTP::Status::status_message($code);
+        my $len = length($resp->content);
+
+        $info{ $stocks, "symbol" } = $stocks;
+
+        if ( $code == 200 && $len > 5 ) {
+            my $payload = substr($resp->content, 5);
+            my $data = JSON::decode_json $payload;
+
+            $info{$stocks, 'success'} = 1;
+            $info{$stocks, 'method'} = 'xetra';
+            $info{$stocks, 'exchange'} = 'XETR';
+            $info{$stocks, 'symbol'} = $data->{'isin'};
+            $info{$stocks, 'last'} = $data->{'lastPrice'};
+            $info{$stocks, 'high'} = $data->{'dayHigh'};
+            $info{$stocks, 'low'} = $data->{'dayLow'};
+            $info{$stocks, 'close'} = $data->{'closingPricePrevTradingDay'};
+            $info{$stocks, 'currency'} = $data->{'currency'}{'originalValue'};
+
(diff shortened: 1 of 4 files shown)


fellen:
I do not see the benefit. According https://code.gnucash.org/docs/C/gnucash-help/fq-spec-yahoo.html you can get it from yahoo_json by appending ".DE" to the ISIN.

e-dorigatti:
Apologies, I was not aware of that page. However, the ISIN in the example above does not seem to work. Is it a problem on my end?

fellen:
https://www.xetra.com/xetra-en/instruments/shares/list-of-tradable-shares/xetra/4750!search?query=DE000A0D8Q07 says
`0 results for "DE000A0D8Q07"`
while "DE0007236101" and "SIE" both return
```
    Shares
    SIEMENS AG NA O.N.

    ISIN: DE0007236101

    CCP eligible

    Market: Xetra
```

e-dorigatti:
Perhaps it does not work because it's an ETF? The ISIN is correct and the quotes are at https://www.boerse-frankfurt.de/en/etf/ishares-euro-stoxx-ucits-etf-de

fellen:
https://www.xetra.com/xetra-en/instruments/etf-exchange-traded-funds/list-of-tradable-etfs/xetra/3044!search?query=DE000A0D8Q07 returns
```
    ETFs
    ISHARES EURO STOXX U.ETF

    ISIN: DE000A0D8Q07

    CCP eligible

    Market: Xetra

```
OK, then your module will also fetch ETFs. That should be mentioned in its docs.

Note: .F (Börse Frankfurt) and .DE (XETRA) are 2 different markets with slightly different quotes. .F is the traditional exchange and .DE its electronic counterpart. Run 
`gnc-fq-dump -v yahoo_json SIE.F SIE.DE`
to compare them.

lib/Finance/Quote/XETRA.pm
```diff
@@ -0,0 +1,128 @@
...
+
+    return wantarray() ? %info : \%info;
+    return \%info;
+}
+
+1;
+
+=head1 NAME
+
+Finance::Quote::XETRA - Obtain quotes from XETRA (Frankfurt Stock Exchange)
```
fellen, line 101: see Note in my previous comment

lib/Finance/Quote/XETRA.pm
```diff
@@ -0,0 +1,128 @@
...
+
+    use Finance::Quote;
+
+    $q = Finance::Quote->new;
+
+    %info = Finance::Quote->fetch("xetra","EU0009658145");
+
+=head1 DESCRIPTION
+
+This module fetches information from the Frankfurt Stock Exchange.
```
fellen, line 113: see Note in my previous comment

fellen (changes requested):
Frankfurt and Xetra are different exchanges.

bpschuck:
Please note, the modules are listed in Modules-README.yml alphabetically. The maintainers also ask that a respective test suite be written and placed in the t/ directory with the other test suites. Include at least one test that should fail. See the other test files for examples.
Thank you.

lib/Finance/Quote/XETRA.pm
```diff
@@ -0,0 +1,128 @@
...
+use vars qw($VERSION $YIND_URL_HEAD $YIND_URL_TAIL);
+use LWP::UserAgent;
+use HTTP::Request::Common;
+use HTML::TableExtract;
+use Time::Piece;
+
+# VERSION
+
+
+my $PRICE_URL = "https://api.boerse-frankfurt.de/data/price_information?mic=XETR&isin=";
```
fellen, line 32: Just thinking loud: If you can set other values for mic. the module could become versatile and allow Frankfurt, XETRA, (more?). You could implement that by suffixes like .F and .X
But then the module name should change, too.

e-dorigatti: Yes, that should work for Frankfurt, using `mic=XFRA`. Not sure about other exchanges and how to find out (did not see other options on the website). Suggestions for a new name?

fellen: I don't know, but you can download the full ISO 10383 list from https://www.iso20022.org/market-identifier-codes and check, which return a result.

If quotes from more mics are available there, I would recommend to use the mic as suffix with "XETR" as default.

fellen: > Suggestions for a new name?

In between I think the name is fine —XETR has much more volume than XFRA—, but the documentation should mention, that and how also XFRA is available.

lib/Finance/Quote/XETRA.pm
```diff
@@ -0,0 +1,128 @@
...
+    use Finance::Quote;
+
+    $q = Finance::Quote->new;
+
+    %info = Finance::Quote->fetch("xetra","EU0009658145");
+
+=head1 DESCRIPTION
+
+This module fetches information from the XETRA exchange, also including ETFs.
+
```
fellen, line 114: Full list of symbols: https://www.xetra.com/xetra-de/instrumente/alle-handelbaren-instrumente

vincentl:
Once the additional line about where to find the full list of symbols is added to the documentation, are we happy with this PR?

e-dorigatti:
I will write tests in a few days, sorry for the delay!

vincentl:
I just merged pull request #149, which changed Quote.pm.  Pull from master and then rebase your branch on the tip of master.

e-dorigatti:
Done, is anything else required?

vincentl:
Thanks!  I'm going to work on merging pull requests this weekend.

Andre-J:
Hi,

is Xetra.pm still experimental or already outdated?

The module is in the Finance/Quote folder, but not listed in quote.pm. If I add it there, Finance::Quote complains that there is no method "xetra" found.

bpschuck:
The URL is no longer valid, thereby breaking the module. It will be removed completely from the next release of F::Q.

fellen:
Yes, there have been changes
https://www.boerse-frankfurt.de/en/relaunch.
@e-dorigatti from where did you get the URL?
If from the android app, can you look for the new URL?

e-dorigatti:
@fellen I got that URL via the web developer tools on their website. But with the new redesign they added countermeasures to prevent this and I cannot figure out how to get the quotes.

bpschuck:
@e-dorigatti - Unfortunately this is not the only module this has happened to over the last few years. If and when you find it can be resurrected, please create another pull request with the update version of the module (and test file).

Thank you for your contributions.

Bruce S.

Andre-J:
So I suggest to remove the "next release" label and add the "module malfunction" label.

bpschuck:
The labels on pull requests and issues are assigned soon after they are created. When created, PR #151 was going into the next release (I think that was 1.50). While labels may change while the issue or PR is open, they are not changed once closed. For a good reference of how it was classified when actively being addressed. There are now quite a few modules that have been added in the past that currently do not work. The F::Q team is not about to go through every past PR and change labels for those modules.

What we have been doing is updating the Modules-README.yml file. Here one can view the state of the module when last tested (yes, some are still TBD), the URL/URI used for fetching the data, and test cases used in the t/mymodule.t test files. You will find that the Modules-README.yml file will get updated a few times between releases.

Thank you.

PR #153, opened 2020-09-29
Changes for new Fundata module.

AbstractMethod:
Creating new Fundata module which pulls Canadian mutual fund information from: http://www.fundata.com


vincentl:
Thanks!

PR #155, opened 2020-10-06
reinstate check for empty return array

jvolkening:
Addresses https://github.com/finance-quote/finance-quote/projects/2#card-46824564


vincentl:
Merged with other changes in 491546cefc39feb4c31930289815744bbd87604b

PR #156, opened 2020-10-06
don't test currency() without AlphaVantage API key

jvolkening:


vincentl (approved):
Ran tests without API_KEY and test is correctly skipped with appropriate message.

PR #158, opened 2020-10-07
Vanguard/AlphaVantage fixes

jvolkening:
It turns out the failure of `vanguard.t` was caused by behavior in AlphaVantage, as the `vanguard()` method is now simply an alias for `alphavantage()`. The `p_change` field was added in a fairly recent commit, but the trailing percent sign was not stripped off. This PR adds code to strip off the trailing percent sign. **I suppose this could be a potentially breaking change if anyone is currently using this value in production**. Surely it's not meant to be there, though?

Also, a test is added for the new `p_change` field, the Vanguard tests are skipped if the AV API key is not present, and a test is removed in `vanguard.t` which checks for a field not provided by AlphaVantage.

### DIFF
--- a/lib/Finance/Quote/AlphaVantage.pm
+++ b/lib/Finance/Quote/AlphaVantage.pm
@@ -247,6 +247,9 @@ sub alphavantage {
         #     }
         # }
 
+        # remove trailing percent sign, if present
+        $quote->{'10. change percent'} =~ s/\%$//;
+
         $info{ $stock, 'success' } = 1;
         $info{ $stock, 'success' }  = 1;
         $info{ $stock, 'symbol' }   = $quote->{'01. symbol'};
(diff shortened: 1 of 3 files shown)


vincentl:
I never realized that vanguard really pointed to AlphaVantage.  

Since we are going to remove modules in release 1.50, what do you thing about removing vanguard altogether since FQ isn't really fetching data from that source?

jvolkening:
> I never realized that vanguard really pointed to AlphaVantage.
> 
> Since we are going to remove modules in release 1.50, what do you thing about removing vanguard altogether since FQ isn't really fetching data from that source?

Makes sense to me. Looks like this was introduced in c61748fb9580dc55320d862b83057cd4453b596a. I don't know the rationale or history behind this.

PR #160, opened 2020-10-07
Data::Dumper and debugging code (was: Misc fixes)

jvolkening:
There is a test specifically to check for `Data::Dumper` being loaded, so these need to be removed. Possibly better just to remove the lines altogether.

### DIFF
--- a/lib/Finance/Quote/Currencies.pm
+++ b/lib/Finance/Quote/Currencies.pm
@@ -24,7 +24,7 @@ use LWP::UserAgent;
 use HTTP::Request::Common;
 use HTML::TableExtract;
 use Encode qw(decode);
-use Data::Dumper::Perltidy;
+#use Data::Dumper::Perltidy;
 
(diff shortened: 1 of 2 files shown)


vincentl:
I favor removing code rather than commenting (unless the comment documents why the active code is doing something unexpected or non-standard).

jvolkening:
Along with the changes here, some other modules also have Data::Dumper code interspersed and commented out, sometimes along with other debugging code. Several modules make use of a hard-coded "DEBUG' constant. What about using something like `Smart::Comments` to replace these? This would clean things up a bit at the cost of an additional dependency.

vincentl:
Smart::Comments looks good to me.

vincentl:
Before merging the PR, I modified it to also use "constant folding" to conditionally turn on Smart::Comments.  I added a description to Documentation/Hackers-Guide as well.  Let me know if you see any issues with what I wrote in the Hackers-Guide (or the strategy in general).  Thanks!

PR #163, opened 2020-10-11
TIAA-CREF fix

jvolkening:
Major re-write of the broken TIAA-CREF module. Greatly simplified method should be stable as long as the backing service URL remains functional. NOTE: some of these securities are available elsewhere and some are not.

Also, added a few testing dependencies that are used here and in other modules to `dist.ini`. They are now in core, but were not for very old perls which apparently are still supported by F::Q.


jvolkening:
Failing due to #164

PR #164, opened 2020-10-11
restore 'code' field in known_currencies()

jvolkening:
I noticed in separate failed PR that the `currency_lookup.t` test unit is now failing, and I think it was due to an earlier PR of mine. Apparently the last fix to `Currencies.pm` fixed one bug but broke expected behavior that was used elsewhere (in Quote.pm `currency_lookup()`). This commit restores the 'code' field so that both 'code' and 'number' are available and `currency_lookup.t` passes.


jvolkening:
Merged into #163

PR #165, opened 2020-10-12
Troweprice fix

jvolkening:
Fixes and updates T. Rowe Price module (broken data source).

I confirmed that all securities returned by this data source are also found in YahooJSON. The direct data source also does not return rich information -- only date and last price. Perhaps it would be a candidate for deprecation.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -646,15 +646,20 @@
     - CREFbond
 #
 - module: Troweprice.pm
-  state: TBD
+  state: working
   added: TBD
-  changed: ~
+  changed: 2020-10-11
   removed: ~
-  urls:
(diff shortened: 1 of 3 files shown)


bpschuck:
On 10/11/20 21:51, Jeremy Volkening wrote:

> Fixes and updates T. Rowe Price module (broken data source).

Awesome, thanks.

> I confirmed that all securities returned by this data source are
> also found in YahooJSON. The direct data source also does not return
> rich information -- only date and last price. Perhaps it would be a
> candidate for deprecation.
That is information perfect for the notes field in Modules-README. For
the long term, may be worth thinking about to put something in a
module's POD code that can get extracted automatically.

- Bruce S.

PR #169, opened 2020-10-19
New xetra url and more resilient reading of data

e-dorigatti:
Hello! Seems like they changed the API endpoint so I updated the URL. I also introduced a callback to read until the first newline since they are not guaranteed to send it all at once.


vincentl:
Thanks!

Issue #166, opened 2020-10-13
ASX source stopped working?


slowbart:
Some time around 11 Oct 2020 my Gnucash install stopped receiving stock quotes from ASX. No changes made to either Gnucash of F::q

Quotes still work from e.g. yahoo_json for the same stock codes. Output below from ASX (failed with **errormsg: Failed to parse HTML table.**) and then Yahoo (working).

Can anyone help please?
```
$ /Applications/Gnucash.app/Contents/Resources/bin/gnc-fq-dump -v asx AFI
Finance::Quote fields Gnucash uses:
    symbol: AFI (deduced)        <=== required
      date: ** missing **        <=== recommended
  currency: ** missing **        <=== required
      last: **missing**          <=\       
       nav: **missing**          <=== one of these
     price: **missing**          <=/        
  timezone:                      <=== optional

** This stock quote cannot be used by GnuCash!


All fields returned by Finance::Quote for stock AFI

stock           field  value
-----           -----  -----
AFI          errormsg: Failed to parse HTML table.
AFI              last: **missing**
AFI               nav: **missing**
AFI             price: **missing**
AFI           success: 0

$ 

$gnc-fq-dump -v yahoo_json AFI.ax
Finance::Quote fields Gnucash uses:
    symbol: AFI.ax               <=== required
      date: 10/13/2020           <=== recommended
  currency: AUD                  <=== required
      last: 6.57                 <=\       
       nav:                      <=== one of these
     price:                      <=/        
  timezone:                      <=== optional

All fields returned by Finance::Quote for stock AFI.ax

stock           field  value
-----           -----  -----
AFI.ax          close: 6.45
AFI.ax       currency: AUD
AFI.ax           date: 10/13/2020
AFI.ax      div_yield: 3.7209302
AFI.ax            eps: 0.199
AFI.ax       exchange: Sourced from Yahoo Finance (as JSON)
AFI.ax           high: 6.57
AFI.ax        isodate: 2020-10-13
AFI.ax           last: 6.57
AFI.ax            low: 6.47
AFI.ax         method: yahoo_json
AFI.ax           name: AFI.ax (AUS.FOUND. FPO)
AFI.ax           open: 6.47
AFI.ax             pe: 33.015076
AFI.ax        success: 1
AFI.ax         symbol: AFI.ax
AFI.ax           type: EQUITY
AFI.ax         volume: 454372
AFI.ax     year_range:         4.77 - 7.32

$
```

[...]

PR #170, opened 2020-10-21
ASX fix

jvolkening:
Fixes the ASX module. Data is now pulled from JSON endpoints.


jvolkening:
Ah, I just saw that some work has already been done on this source in #166. Feel free to close this as a duplicate.

CleanShed:
Hi @jvolkening , I did some work on #166 and wasn't aware of your solution here.  Good job identifying that fallback URL.

I've tested and found a couple of issues.  How should we proceed?  I'm a newbie and happy to let you run with this. 
 Are you able to rectify these issues?  Otherwise I can fold your solution back into #166   Thanks

(1) ASX.pm returns yesterday's date.  
[fq_asx_170_dates.txt](https://github.com/finance-quote/finance-quote/files/5422179/fq_asx_170_dates.txt)

(2) ASX.pm sometimes returns undefined values, causing problems for GunCash.
[fq_asx_170_undef.txt](https://github.com/finance-quote/finance-quote/files/5422222/fq_asx_170_undef.txt)

jvolkening:
@CleanShed -- feel free to proceed with your solution; we must have been working on it at the same time. Regarding your issues mentioned:

> (1) ASX.pm returns yesterday's date.

The `last_trade_date` field returned by the endpoint is converted to a `Time::Piece` object and then converted to a local timestamp. Since the original times are UTC+11, it ends up as the previous day in many parts of the world. I actually don't know what the expected behavior is here in terms of conversion to local time or not.

> (2) ASX.pm sometimes returns undefined values, causing problems for GunCash.

I think there are two issues here:

1. You identified another edge case that my code didn't handle: namely, the call to the main endpoint returns a 200 code and a JSON response but the response does not contain any useful data. Another conditional is needed to test that the response contains the expected fields. 

2. For securities which are not available via the first endpoint and rely on the second, the response will not have as many labels/keywords defined. I don't know how this behavior affects downstream consumers such as GnuCash. If this is an issue, I can only see two solutions: (1) reduce the number of labels advertised as available by the module to the least common subset of the two data sources, or (2) don't return data from the fallback endpoint and simply report these securities as unavailable via ASX.pm.

jvolkening:
@CleanShed I pushed a commit to handle the additional edge case you mentioned...BUT, if you are actually using this module, you should proceed with your solution and submit a PR, and then I will close this one. I was simply working on it as part of addressing all of the outstanding test failures.

Also, a major shortcoming of using the fallback data source is that it does not return a date associated with the data. On the one hand, this seems like a significant problem and argues against using this data. On the other hand, some of the securities available via this endpoint are not available through other major data providers such as Yahoo JSON.

goodvibes2:
If F::Q does not supply a date, GnuCash script gnc-fq-helper, thru which all F::Q calls are made, will supply the current date, so as far as GnuCash is concerned, it is acceptable for the date to be missing from the data returned. Thanks for working on this.

jvolkening:
@goodvibes2 Thanks for the explanation. Can you advise on whether F::Q should return the quote date in user local time or verbatim as provided by the exchange?

goodvibes2:
@jvolkening,
As far as I know, date as supplied by ASX is fine. GnuCash only allows 1 price database record per stock/day and ASX has only fairly recently been able to supply a date. I don't think it really matters as this is only used for an estimated portfolio value in reports. By the way how did you find out about this ASX json web data? Are there any restrictions on its use?

jvolkening:
@goodvibes2 I traced the JSON endpoints simply watching the XHR network traffic in the browser. As far as restrictions, I'm not sure. There is the typical clause in the ASX website terms of use:

> You agree not to use any spider, screen scraper, robot, or other automated similar software or device to use or access the site in any way whatsoever, including monitoring, downloading or copying any of the content on the Site

We're not actually screen scraping here (like the previous version of the module did, and many other F::Q modules do), so I'm not sure where the legal balance lies.

CleanShed:
> @CleanShed I pushed a commit to handle the additional edge case you mentioned...BUT, if you are actually using this module, you should proceed with your solution and submit a PR, and then I will close this one. I was simply working on it as part of addressing all of the outstanding test failures.
> 
Hi @jvolkening  Yes I do use ASX.pm so I will move your solution across to #166 , test it, and fix any issues including dates etc as best I can, then submit a PR.  Thanks very much for your efforts!

The fallback URL seems to be useful for certain hybrid securities, so I am thinking of calling it to augment the missing data from the primary URL if I can see a useful pattern.

CleanShed:
> @jvolkening,
> As far as I know, date as supplied by ASX is fine. GnuCash only allows 1 price database record per stock/day and ASX has only fairly recently been able to supply a date. I don't think it really matters as this is only used for an estimated portfolio value in reports.

The documentation in Quote.pm says:
`    date         Last Trade Date  (MM/DD/YY format)`
`    time         Last Trade Time`
but has no concept of timezones.  The current version of ASX.pm returns neither date nor time.  In my opinion, the new version should return them only if supplied by the data sources, and "as supplied" ie in the local timezone of the exchange.

goodvibes2:
@jvolkening, I've never seen that terms of use clause before. I'm no lawyer but it seems to me ASX does not allow F::Q to use json or screen scraping. Anyone else know anything to the contrary?

CleanShed:
> @goodvibes2 I traced the JSON endpoints simply watching the XHR network traffic in the browser. As far as restrictions, I'm not sure. There is the typical clause in the ASX website terms of use:
> 
> > You agree not to use any spider, screen scraper, robot, or other automated similar software or device to use or access the site in any way whatsoever, including monitoring, downloading or copying any of the content on the Site
> 
> We're not actually screen scraping here (like the previous version of the module did, and many other F::Q modules do), so I'm not sure where the legal balance lies.

@goodvibes2  this is interesting.  The [terms of use](https://www2.asx.com.au/legals/terms-of-use) document actually defines the Site as:
`www2.asx.com.au is a website (the “Site”) owned and operated by ASX Operations Pty Limited`
however the JSON data sources are both from different URLs and do not appear to have any terms of use...

frankcrawford:
> > (1) ASX.pm returns yesterday's date.
> 
> The `last_trade_date` field returned by the endpoint is converted to a `Time::Piece` object and then converted to a local timestamp. Since the original times are UTC+11, it ends up as the previous day in many parts of the world. I actually don't know what the expected behavior is here in terms of conversion to local time or not.

Yes, I've been trying this out, and this particular item is the only thing causing me any concern.  As last_trade_date is really a date and not a timestamp, I don't believe you should try and interpret the timezone.  What is worse is that the time is set to 00:00, and hence will go backwards for most parts of the world.  What is really needed is to put in the local timezone, when generating the dates, but I can't see how to do that, or otherwise it may be best just to cut off that part of the entry.

vincentl:
I suggest just extracting the 'YYYY-MM-DD' part of last_trade_date and pass it to 
`$quoter->store_date(\%info, $stocks, {isodate => 'YYYY-MM-DD'});`.

frankcrawford:
So what you are suggesting is that it should be:

    `my $date = $date->{last_trade_date};`
    `$date =~ s/T..:..:.......$//;`
    `$quoter->store_date(\%info, $stocks, {isodate => $date});`

(which could be all shrunk down).

vincentl:
That code should work.  If I'm feeling particularly paranoid, I might write it like
`$quoter->store_date(\%info, $stocks, {isodate => $1}) if exists $data->{last_trade_date} and $data->{last_trade_date} =~ /([0-9]{4}-[0-9]{2}-[0-9]{2})T/;`;

If the json changes, then `$data->{last_trade_date}` won't exists or the regular expression capture will fail and store_date() won't be called.  Then the quote will be returned without a date and the caller can decide was to do (most likely reject the quote).

PR #172, opened 2021-01-14
ASX Oct-2020 web site fix

CleanShed:
Updated ASX.pm and asx.t to cater for breaking changes made by the ASX in October 2020.

lib/Finance/Quote/ASX.pm
  October 2020 the ASX revamped their website with dynamic content for quotes
  which prevented the previous HTML screen scraping from working, but exposed
  a number of JSON data sources, two of which are used here.
  The primary source returns data elements for almost all securities, but
  does not return prices for certain security types (some bonds and exchange
  traded products, options, and warrants), and returns an error for indices.
  The alternate source returns less data elements, but provides usable quote
  data for all the known exceptions, including indices.
  This version will always call the primary source, and call the alternate if
  a price is not returned by the primary.
  
  Smart::Comments implemented to conform with the Hackers Guide:
  	https://github.com/finance-quote/finance-quote/blob/master/Documentation/Hackers-Guide

t/asx.t
  Update the list of securities to test against.
  Include additional types of securities:
    Corporate Bonds
    Government Bonds
    Exchange Traded Products (eg ETFs)
  Refactor into sub tests by security type.
  Update the data elements to be tested.
  Skip data element tests when security not found.


CleanShed:
RE: "continuous-integration/travis-ci/pr — The Travis CI build failed" errors

All those errors are in modules that I haven't touched - how do I proceed from here?

Thanks

fellen:
@CleanShed CleanShed, you could try to rebase on current head and if that succeeds, force push.

vincentl:
Don't worry about the other modules.  Thanks so much!!  I merged your changes into my master and all your tests pass.  I'm down to only 2 modules to fix and then all test should pass.

PR #176, opened 2021-03-07
fix some POD issues

gregoa:
In Debian we are currently applying the following patch to
Finance-Quote.
We thought you might be interested in it too.

    Description: fix some POD issues
    Origin: vendor
    Author: gregor herrmann <gregoa@debian.org>
    Last-Update: 2021-03-06
    

The patch is tracked in our Git repository at
https://salsa.debian.org/perl-team/modules/packages/libfinance-quote-perl/raw/master/debian/patches/pod.patch

Thanks for considering,
  gregor herrmann,
  Debian Perl Group

### DIFF
--- a/lib/Finance/Quote/ASX.pm
+++ b/lib/Finance/Quote/ASX.pm
@@ -326,32 +326,45 @@ sub process_asx_data {
 __END__
 
 =head1 NAME
+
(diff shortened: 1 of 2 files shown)


vincentl:
Thanks for these updates

PR #201, opened 2022-03-05
Maintainers prefer issues get reported at github

jayenashar:
https://rt.cpan.org/Transaction/Display.html?id=2197790

> The maintainers prefer that issues for F::Q get reported at https://github.com/finance-quote/finance-quote instead of cpan.org.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -85,8 +85,8 @@ repository.web = https://github.com/finance-quote/finance-quote
 repository.type = git
 homepage = http://finance-quote.sourceforge.net/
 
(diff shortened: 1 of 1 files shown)


bpschuck:
Thanks. I was going to put this in the 1.52 project plan as a cleanup item. I also need to fix some broken URLs on the old SourceForge site (htdocs/developer.html).

bpschuck:
For the time being we will continue to use CPAN's RT Bug Tracking system. We need to go through old bug reports and clean them out.

PR #206, opened 2022-05-06
Changes to ZA module

thinuspollard:
Updated Sharenet parsing
    
    Sharenet reports share prices in minor denominations. All other examples
    reports in major denominations. GnuCash specifically is expecting major
    denomination.
    
    Added the following fields:
    * Exchange: JSE
    * Last: set to the same as price
    * Source: set to sharenet.co.za
    * Symbol: set to the input symbol

The only concern I have is the change in the interface in terms of major/minor denominations that may break other applications. According the wiki page only GnuCash is using this.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -848,13 +848,14 @@
   testcases:
 #
 - module: ZA.pm
-  state: TBD
-  added: TBD
-  changed: ~
+  state: working
+  added: 
+  changed: 2022-05-10
   removed: ~
-  urls:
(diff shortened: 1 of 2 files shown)


bpschuck:
Thank you for your contribution. It would be a huge help to the maintainers if you could also update Modules-README.yml. Look at the entries for AlphaVantage.pm for an example of the information to be noted.
Thanks!

thinuspollard:
Ok, will update and issue a new pull request

PR #207, opened 2022-05-09
Re-enable morningstar.ch.

fritteli:
Please include morningstar.ch again - all that's needed is the HTTPS URLs instead of the HTTP URLs.


bpschuck:
Thank you for your contribution. It would be a huge help to the maintainers if you could also update Modules-README.yml. Look at the entries for AlphaVantage.pm for an example of the information to be noted.
Thanks!

fritteli:
Sure - thanks for the hint. I will update the file as soon as I have time, within the next few days, hopefully.

fritteli:
I updated the Modules-README.yml

bpschuck:
Thanks!

PR #208, opened 2022-05-10
Changes to ZA module

thinuspollard:
Updated Sharenet parsing

Sharenet reports share prices in minor denominations. All other examples
reports in major denominations. GnuCash specifically is expecting major
denomination.

Added the following fields:
* Exchange: JSE
* Last: set to the same as price
* Source: set to sharenet.co.za
* Symbol: set to the input symbol

The only concern I have is the change in the interface in terms of major/minor denominations that may break other applications. According the wiki page only GnuCash is using this.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -848,13 +848,14 @@
   testcases:
 #
 - module: ZA.pm
-  state: TBD
-  added: TBD
-  changed: ~
+  state: working
+  added: 
+  changed: 2022-05-10
   removed: ~
-  urls:
+  urls: https://www.sharenet.co.za/jse/$symbol
   apikey: false
-  notes: ~
+  notes: 
+    Module to acquire data from https://www.sharenet.co.za
   testfile: TBD
   testcases:
 #
--- a/lib/Finance/Quote/ZA.pm
+++ b/lib/Finance/Quote/ZA.pm
@@ -120,7 +120,7 @@ Finance::Quote->new().
 
 =head1 LABELS RETURNED
 
-The following labels will be returned: success currency name price date isodate symbol.
+The following labels will be returned: success currency name price date isodate symbol last source exchange.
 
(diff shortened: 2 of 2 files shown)


bpschuck:
Thanks. BTW - If you speaking about the GitHub Wiki page, it was never really put to use and updated. So while GnuCash is the only user noted in the Wiki here, I would not assume there are no others.

thinuspollard:
Yeah, that's what I figured. Just noting it so that other users maybe see it.

I guess I can enable both behaviors based on a parameter, with the current behavior as default if the parameter is not present. And then get GnuCash to add support for this.

I don't really know what the right way forward is here. This change makes a breaking change to the interface, but I have no idea how many people are using this at the moment.

bpschuck:
I'll merge it sometime soon. Need to do some housekeeping here and push a 1.52 release. Hopefully in June.

bpschuck:
Hello,
I was going to merge this PR today but double checked. The change to the Modules-README.yml file is incomplete. The name of the testfile is missing. While it should be za.t, I then would ask is if the testfile as is (modified over a year ago) has been tested? Do the securities used as test cases still return valid data. If so, they would also be noted in the Modules-README.yml file.
Thank you.

bpschuck:
Trying to execute the test it produced the error:
`bruce.schuck@Bruces-MacBook-Pro-2 t % perl za.t
Argument "2/8" isn't numeric in multiplication (*) at za.t line 44.
Argument "4/8" isn't numeric in multiplication (*) at za.t line 44.`

My preference would be to change line 44 to
`plan tests => 1 + (scalar keys %check) * (scalar keys %valid) + scalar @invalid;`

In the ZA.pm stanza in the Modules-README.yml file you should have:
```testfile: za.t
testcases:
  - AGL
  - AMS
```
Thanks.

PR #243, opened 2022-10-18
New modules Sinvestor.pm Tradegate.pm and XETRA.pm

Andre-J:
In order to use ISIN symbols for my investments, I added some modules for retrieving prices using the investment portal of the German Sparkasse Group called S-investor.

It delivers prices for shares and ETF that are traded on German and some European online and physical exchanges. Furthermore, fund prices from the investment companies are retrieved as well, supplied by Lipper.

The S-investor system makes a decision on its own what source to take, but the exchange is returned by my module. You can however specify a dedicated exchange by adding a dot and exchange symbol to the ISIN. Possible values are:
TDG Tradegate
GTX Tettex
0QT Quotrix
GER XETRA
FRA Frankfurt
MUN München
BER Berlin
STU Stuttgart

Tradegate.pm and XETRA.pm are spin-offs of Sinvestor to limit the call to the two online trading platforms with the most trading volume. The main difference is that XETRA closes at 17:00 local time with the physical exchange in Frankfurt, and the DAX index is calculated from that.

Please test the modules. I will add some CPAN-compliant test files later.


bpschuck:
You are assuming that I would know what symbols to test...

bpschuck:
FYI - The README.md file also lists the modules in the "See Also" section. The `VERSION` line in the modules should be `# VERSION` (the version number gets set automatically during the Dist::Zilla build and release process).

bpschuck:
WRT a timeline for v1.54, TBD.

Andre-J:
I included testfiles and the other stuff you mentioned.

bpschuck:
Thanks for that. I'd prefer more than one test (not including the bogus symbol expected to fail) symbol. Most of the other test files contain at least 2 or 3 valid symbols. But, nevermind. Thanks for the input.

Andre-J:
I tried a similar test module as aex.t.
 It turns out that my modules can only handle one symbol per call. So I have expanded my test case with three calls one after another.

PR #250, opened 2022-11-21
Update for Fondsweb.pm

Andre-J:
Fondsweb.pm failed in GnuCash because the ISIN was not retrieved correctly.
When both ISIN and WKN are present on the website, the module concatenated them. GnuCash expects the matched symbol returned, so I split the value after the 12th position.

Furthermore, I added some more test cases to fondsweb.t, and Modules-README.yml.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -283,6 +283,22 @@
     - DE0008474511
     - LU0051755006
 #
+- module: Fondsweb.pm
+  state: working
+  added: TBD
+  changed: 2022-11-21
+  removed: ~
+  urls:
+  apikey: false
+  notes: ~
+  testfile: fondsweb.t
+  testcases:
+    - LU0804734787
+    - DE0008491002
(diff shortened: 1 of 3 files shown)


bpschuck:
Thank you for your contribution. Could I bother you to add the `https://www.fondsweb.com/de/`  URL to the Fondsweb.pm section you added to `Modules-README.yml`? Thanks.

Andre-J:
Done.
It might be worth sorting that file alphabetically. Maybe you have tools that could do that.

bpschuck:
It is (or at least was) sorted alphabetically. But it was also sorted 
case sensitive ascii. Upper case letters came before lower case, which 
would mean a module named ZZZ.pm would come before AlphaVantage.pm.

I haven't yet looked for a tool to sort YAML. At one time the team was 
looking into that and a way to generate the Modules-README.yml file.

Issue #261, opened 2023-03-29
www.morningstar.co.jp was replaced by www.wealthadvisor.co.jp


afukumoto:
www.morningstar.co.jp stopped the service, and www.wealthadvisor.co.jp provides the same service since March 30.
I could fetch the fund quotes from the new site by this patch:

```
--- MorningstarJP.pm.orig       2022-10-09 04:27:00.000000000 +0900
+++ MorningstarJP.pm    2023-03-30 00:57:10.431992687 +0900
@@ -32,7 +32,7 @@

 # NAV information (basis price)
 $MORNINGSTAR_JP_URL =
-  ('https://www.morningstar.co.jp/FundData/DownloadStdYmd.do?fnc=');
+  ('https://www.wealthadvisor.co.jp/FundData/DownloadStdYmd.do?fnc=');

 sub methods { return ( morningstarjp => \&morningstarjp ); }
 sub labels  { return ( morningstarjp => [qw/symbol date nav/] ); }
```

Although, I guess all reference of `morningstarjp`  should be replaced with `wealthadvisorjp` ?

PR #263, opened 2023-04-08
URL change in MorningstarJP.pm

bpschuck:
URL fix noted in Issue #261.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* URL Change for MorningstarJP. Issue #261
 
 1.54      2022-12-26 15:25:02-08:00 America/Los_Angeles
 	* Fix to AEX.pm - Issue #235 - PR #244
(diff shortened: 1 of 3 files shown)


afukumoto:
In case it matters, Morningstar has no relation to www.wealthadvisor.co.jp anymore, if I understand correctly.
Formerly, Morningstar brand in Japan was licensed to SBI Global Asset Management, but the brand is now licensed to Ibbotson Associates Japan.  Wealth Advisor is a subsidiary of SBI Global Asset Management.

This patch to the `MorningstarJP.pm` works for the users from previous versions with minimum modification, but It might be confusing to future new users if `MorningstarJP.pm` links to www.wealthadvisor.co.jp

bpschuck:
I thought about that. For now I am leaving the module name and method as is for fewer code changes. At some point the module will be renamed to WealthadvisorJP.pm as well as rename the method, but still have both morningstarjp and wealthadvisorjp in the methods.
Thanks.

PR #280, opened 2023-05-22
Fix for Issue 278

vincentl:
Added test case for space in quote value and added substitution operators to squash anything but numbers and period '.' in anything that has a money value.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,6 @@
 {{$NEXT}}
-  * Added get_features method - PR #260
+	* Added get_features method - PR #260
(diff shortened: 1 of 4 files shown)


bpschuck (changes requested):
@vincentl - Thanks. But if you could please update Modules-README.yml and Changes that would be appreciated. Cuts down on housekeeping that needs to be done before pushing a release to CPAN.

vincentl:
Changes and Modules-README.yml updated

bpschuck (approved):

PR #285, opened 2023-05-26
New module - TwelveData.pm

vincentl:
Added TwelveData.pm for quotes.  Updated Changes and Modules-README.yml.

In development tree, test with
```bash
$ ONLINE_TEST=1 TEST_TWELVEDATA_API_KEY=xxx...xx dzil run prove -lv t/twelvedata.t
```

Register at https://twelvedata.com to get API key.

- Free tier limits rate to 8 quotes per minute.  This is built into the module - but if someone subscribes we don't have a way to remove rate limit.
- TwelveData can also do currency conversion, but companion Currency module not written

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,10 @@
 {{$NEXT}}
-    * Added get_features method - PR #260
-    * Added new TwelveData module
+	* Added new TwelveData module
+	* Updated YahooJSON.pm to use https://query2.finance.yahoo.com/v11 - PR #284
+	* Updated CurrencyRates/YahooJSON.pm to use https://query2.finance.yahoo.com/v11 - PR #284
+	* Bourso.pm - Squash anything but numbers and period in quote values.
+	* Renamed MStarUK.pm to MorningstarUK.pm
+	* Added get_features method - PR #260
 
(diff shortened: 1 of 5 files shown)


bpschuck:
@vincentl - Do not die if there is no API Key defined. Look at how AlphaVantage handles it.

```
if ( !defined $token ) {
    $info{ $stock, 'success' } = 0;
    $info{ $stock, 'errormsg' } =
            'An AlphaVantage API is required. Get an API key at https://www.alphavantage.co';
    next;
}
```

Also, going forward I think it would be helpful to add information about what the Free APIs allow in terms of throttling and number of requests per day/month etc in the Pod documentation.

While you are making the change, if you could also merge recent changes into this branch to eliminate conflicts to resolve when merging the pull request, that would be appreciated.

Thank you.

bpschuck:
I may change the key in twelvedata.t to use the regular `TWELVEDATA_API_KEY` environment variable. Imagine if while testing a bunch of the modules you had to basically duplicate already set environment variables. 

While the throttling works, I prefer how it was done in AlphaVantage. At least this way, if one was just fetching data for a handful of securities, they get returned quicker.

Thanks.

PR #294, opened 2023-05-31
Add GitHub Actions

### LINKED ISSUE #293: Add GitHub Actions
to make sure all the tests pass on all the supported platforms on all the supported versions of perl.

szabgab:
the various environment variables are now turned off as for now they cause some test to fail.

closes #293

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,56 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+#  schedule:
+#    - cron: '42 5 * * *'
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu-latest]
+        perl: [ '5.36' ]
+        #runner: [ubuntu-latest, macos-latest, windows-latest]
+        #perl: [ '5.30', '5.36' ]
+        exclude:
+          - runner: windows-latest
+            perl: '5.36'
+
+    runs-on: ${{matrix.runner}}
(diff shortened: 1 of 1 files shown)


bpschuck:
I am also thinking that for F::Q, the action only needs to be done when a pull request is merged or pushed just to the master branch. Even then, perhaps just when merged. Very seldom is a change made directly to the master branch, and typically those are very minor and usually something like updating `Changes` or the htdocs/index.html file that gets uploaded to SourceForge.
Thanks.

szabgab:
Ideally you'd run the tests on every change on every branch to make sure that this change did not break anything. However minor that change might be, it can have a negative effect on some other place. 

e.g. Moving the "use strict" statement uncovered a test failure of having "use Data::Dumper".  True it was not a real bug, but it illustrates that even small changes can affect things that are seemingly unrelated.

It is especially true on code-base that is relatively old.

bpschuck:
On 5/31/23 9:23 PM, Gábor Szabó wrote:

> Ideally you'd run the tests on every change on every branch to make sure 
> that this change did not break anything. However minor that change might 
> be, it can have a negative effect on some other place.

> e.g. Moving the "use strict" statement uncovered a test failure of 
> having "use Data::Dumper". True it was not a real bug, but it 
> illustrates that even small changes can affect things that are seemingly 
> unrelated.

> It is especially true on code-base that is relatively old.

Almost all the branches are going to be deleted. Some of them are years 
old for modules that use sources that are now non-existent.

So, you will not convince me otherwise. The only branch that would need 
to be run through CI is the master branch after PRs are merged.

szabgab:
If there are no changes in a branch then the CI won't run, but ultimately you can configure which branch to even monitor.

PR #296, opened 2023-06-01
Initial version of YahooWeb module

vincentl:
Pulling data from a Yahoo Finance Web page is delicate.  Here is an initial implementation, but there are questions/limitations:

- What does "last" mean?  I have it pulling the value that appears on the left side under the security name, which is the current price during the day and close price in the evening/weekend.  Should last really be the extending trading price shown on the right side?
- Some of the values are not contained in descriptive HTML tags, so there is best effort to capture the correct value, but it is fragile.  However, there is no validation in the current implementation, so we could return garbage without error.
- The date is set to "day of request" and not pulled from the page.  During the trading week (at least) there is no obvious date field to use.  Not sure behavior on the weekends or holidays.

I recommend we do some testing and consistency checking before merging.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -1022,6 +1022,33 @@
     - AENA.MC
     - CFR.JO
 #
+- module: YahooWeb.pm
+  state: working
+  added: 2023-05-31
+  changed:
+  removed:
+  urls:
+    - https://finance.yahoo.com/quote/{$symbol}?p={$symbol}
+  apikey: false
+  notes: ~
+  testfile: yahooweb.t
+  testcases:
+    - MSFT
+    - AMZN
+    - AAPL
+    - GOOGL
+    - GOOG
+    - META
+    - CSCO
+    - INTC
+    - CMCSA
+    - PEP
+    - BRK-A
+    - SEB
+    - NVR
+    - BKNG
+    - IBKR
+#
 - module: YahooYQL.pm
   state: TBD
   added: TBD
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -103,6 +103,7 @@ use vars qw/@ISA @EXPORT @EXPORT_OK @EXPORT_TAGS
     Union
     XETRA
     YahooJSON
+    YahooWeb
     ZA
 /;
 
@@ -1708,6 +1709,7 @@ http://www.gnucash.org/
   Finance::Quote::TwelveData,
   Finance::Quote::Union,
   Finance::Quote::YahooJSON,
+  Finance::Quote::YahooWeb,
   Finance::Quote::ZA
 
 You should have received the Finance::Quote hacker's guide with this package.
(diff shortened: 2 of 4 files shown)


bpschuck:
Don't forget the `Changes` file. Most recent changes near the to please.

I've also realized that the examples in many of the POD sections is incorrect.

```
 use Finance::Quote;
 $q = Finance::Quote->new('YahooWeb');
 %info = Finance::Quote->fetch('yahooweb', "IBM", "AAPL");
```
should be
```
 use Finance::Quote;
 $q = Finance::Quote->new('YahooWeb');
 %info = $q->fetch('yahooweb', "IBM", "AAPL");
```
Last, price, and close have been used somewhat interchangeably looking at some of the older modules. But I would do the same and pick the same value. For the date, might be interesting to look at page over the weekend and see what "At close: 04:00PM EDT" changes to on days the market is closed. 

In BVB.pm I used a combination of HTML::TreeBuilder and HTML::TableExtract. I used TreeBuilder to find the table, the TableExtract to get the data from the table. Maybe look at doing something like that if possible. Looks like "previous close", "open", "EPS", and a few others are in easy to grab Table Data tags. 

`$eps = $tree->look_down(_tag => 'td', data-test => 'EPS_RATIO-value')->as_text` 

Hopefully Yahoo doesn't block the currently used URL too soon. This doesn't need to be rush yet, so it can be tweaked a bit before merging.

Thanks!

bpschuck:
@vincentl - From what I see today, it appears that if it's not a trading day the time under the price changes to something like "At close: June 2 04:00PM EDT". There is a normal trading close and an after hours value and date.

vincentl:
@bpschuck - I captured the raw page with

```bash
curl --user-agent 'Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/81.0' 'https://finance.yahoo.com/quote/AAPL?p=AAPL' 2>&1 | tee log-saturday.txt
```
and the date is filled in by some javascript code!  The raw HTML just has 
```html
<span>At close:   04:00PM EDT</span>
```

bpschuck:
Fudge... I didn't look at the raw HTML. Nevermind.
On a related note. Do you care to look at AEX.pm? It also no longer gets the trade date, which causes aex.t to fail if online testing is enabled. If yes, open an issue and assign it to yourself. OTW I'll get to that maybe in the next week or so. I _still_ have some family/estate and hurricane issues I'm contending with.

bpschuck:
@vincentl  - Any further thoughts on the missing date yet. I've started looking at Date::Manip::Holidays and Date::Business. Maybe pre-seed either with the next 3 or 4 years of holidays and provide a function to all modules that if you call it with a date (usually the current date) it returns the isodate/date of most recent non-holiday weekday. 
So if I call it with July 3, 2023 it returns 2023-07-03. Give it July 4, 2023 and also returns 2023-07-03. Call it with a date that is a Sat or Sun, returns the date of the previous Friday (unless of course it's a holiday, in which case the isodate of the previous Thursday).
Just thinking out loud here.

bpschuck:
While it's the aftermarket trade time, the time in epoch ca be found in:

```
<fin-streamer data-symbol="PDM" data-field="postMarketTime" data-trend="none" value="1686946498"
active="">04:14PM EDT</fin-streamer>
```
Will need to check this value during weekends.

PR #300, opened 2023-06-06
IndiaMutual.pm: TMP and TMPDIR env variables do not include trailing slashes in all distributions

vishalg:
The location of the cache-file is constructed using TMPDIR and TEMP environment variables and these are not not guaranteed to include a trailing slash in all setups. The location of the amfinavlist.txt cache was changed such that a slash is prefixed to the file name

### DIFF
--- a/lib/Finance/Quote/IndiaMutual.pm
+++ b/lib/Finance/Quote/IndiaMutual.pm
@@ -24,8 +24,8 @@ $AMFI_MAIN_URL = ("http://www.amfiindia.com/");
 $AMFI_URL = ("https://www.amfiindia.com/spages/NAVAll.txt");
 
 # amfinavlist.txt is a cache-file. keep it until updated on the website since this is a 1meg file.
-my $cachedir = $ENV{TMPDIR} // $ENV{TEMP} // '/tmp/';
-$AMFI_NAV_LIST = $cachedir."amfinavlist.txt";
+my $cachedir = $ENV{TMPDIR} // $ENV{TEMP} // '/tmp';
+$AMFI_NAV_LIST = $cachedir."/amfinavlist.txt";
 
 sub methods { return (indiamutual => \&amfiindia,
                       amfiindia => \&amfiindia); }


vishalg:
Hi, Thanks for looking at this. I am not the original author but as a best guess

- $TEMP and $TMPDIR have default settings on Debian and the current version will likely not work with the OS defaults. 
- The tests in indiamutual.t seem to be valid but they need ONLINE_TEST variable to be set, does that mean maybe that the tests are all skipped unless the variable is set?
- The file being downloaded is large (~1.5Mb) and the data is updated only once a day. This could be the reason why the original author decided to keep a cache

bpschuck:
- > $TEMP and $TMPDIR have default settings on Debian and the current version will likely not work with the OS defaults.

I beg to differ, none of the last 3 or 4 Ubuntu LTS releases (a Debian derivative as you likely know) set `$TEMP` or `$TMPDIR`. One should never presuppose that an environment variable is already set. Windows and Mac users are also users of F::Q. Processing the file from memory instead of a disk file would make this a non-issue.

- > The tests in indiamutual.t seem to be valid but they need ONLINE_TEST variable to be set, does that mean maybe that the tests are all skipped unless the variable is set?

Correct. Most people don't have it set, but modules should be tested _before_ submitting pull requests.

- > The file being downloaded is large (~1.5Mb) and the data is updated only once a day. This could be the reason why the original author decided to keep a cache.

1.5Mb is not really _large_ anymore. The author may not have realized that the data could just as easily be processed from memory as well.

I'm also not the original author of just about all the modules in F::Q that I've need to tweak in the last few years since I've been sort of the lead maintainer. That's not uncommon for community supported OpenSource software.

Thanks.

vishalg:
Noted. I will close the pull request as the wider fixes are beyond my competence.

bpschuck:
Actually it was trivial.  And one test case in `indiamutual.t` was no longer traded so I changed it. See https://github.com/finance-quote/finance-quote/pull/301

PR #301, opened 2023-06-06
Changed IndiaMutual.pm to use IO::String

bpschuck:
Changed IndiaMutual.pm to use IO::String object instead of downloading and parsing a file. Eliminates issues caused by incorrect or non-existing directories. Changed a non-working test case. Updated Changes and Modules-README.yml files.


bpschuck:
I need to make some minor changes to the POD section before merging.

vishalg:
That's great. Much appreciated. Sorry I couldn't help as I haven't written in Perl for about last 15 years.

PR #304, opened 2023-06-18
sinvestor pages enhanced by close, change, p_change

mumpitzstuff:
- close, change and p_change added
- bug fixed for values above 999.9 (1.000,0 converted to 1000.0)

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,6 @@
 {{$NEXT}}
+	* Added close, change and p_change to Tradegate, XETRA and Sinvestor
+	  Fixed bug in Tradegate, XETRA and Sinvestor for numbers equal or higher than 1.000 - PR #304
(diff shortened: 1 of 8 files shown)


bpschuck:
Thank for for the contribution. 
If it's not too much of a bother, could you please confirm the symbols used in the test file (`t/sinvestor.t`) are all still valid and update if any are no longer traded?
Another thing the team asks is to update the `Modules-README.yml` file, and if possible the `Changes` file (in this file recent changes should be in the top of the list directly underneath `{{$NEXT}}`.
This helps with housekeeping done just before pushing a release to CPAN.
Thank you!

mumpitzstuff:
- testcases updated
- % removed from p_change because i have seen in other testcases that this is normally not part of the return value
- modules readme updated
- changes updated

PR #308, opened 2023-06-26
add INST_ID as optional parameter to XETRA, Tradegate and Sinvestor

mumpitzstuff:
Behaviour is exactly the same if no INST_ID is provided as parameter. This means the user can use the institute id but does not have to.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 {{$NEXT}}
 	* Correct set exchange in YahooJSON.pm - Issue #306
 	* Added close, change and p_change to Tradegate, XETRA and Sinvestor
+	  Added optional parameter INST_ID to specify the institute id.
 	  Fixed bug in Tradegate, XETRA and Sinvestor for numbers equal or higher than 1.000 - PR #304
 	* Added GoogleWeb Module
 	* YahooWeb module added - PR #296
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -712,9 +712,13 @@
 - module: Sinvestor.pm
   state: working
   added: 2022-10-19
-  changed: 2023-06-18
+  changed: 2023-06-26
   removed: ~
-  urls: https://web.s-investor.de/app/aktien.htm?INST_ID=0000057
+  urls:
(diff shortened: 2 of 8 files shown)


bpschuck:
I'm thinking this would be easier to just map the INST_ID to the method. Much like many modules may map 'nyse' to their native method such as yahooweb or alphavantage.

mumpitzstuff:
Hmm i really have no idea how it should work. I do not want to provide hardcoded institute ids because they may change in the future. Therefore i would prefer an optional parameter to let the users decide if they want to use the default institute (SParkasse Krefeld) or any other institute like Sparkasse München).
If you really think that a method should be used please provide a pseudocode example how such a call would look like.

bpschuck:
Ignore that.
Does the error message returned in `$@` contain useful information if the fetch was called with a non-existing INST_ID?

bpschuck:
@mumpitzstuff 
I wrote a quick test script to see what the error may contain if given an invalid INST_ID.

```
#!/usr/bin/perl
# vi: set ts=2 sw=2 noai ic showmode showmatch: 

use Finance::Quote;

my $q = Finance::Quote->new('XETRA', 'xetra' => {INST_ID => '0099903'});

my %info = $q->fetch('xetra', 'NL0000009165');

foreach $key (keys %info) {
  $value = $info{$key};
  print "$key: $value\n";
}
```

When executed it returns:

```
NL0000009165errormsg: Error retreiving NL0000009165: Can't call method "look_down" on an undefined value at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/XETRA.pm line 75.

NL0000009165success: 0
```

In my opinion it would be more useful to catch the error from the initial fetch, or in this case that the initial call to TreeBuilder failed because the URL contained a non-existent/invalid INST_ID.

mumpitzstuff:
Its a valid finding. I added something to catch such an error...

bpschuck:
Cool, thanks. I see you also added where one can find a list of the INST_ID to the various PODs. Thanks for that detail.

PR #331, opened 2023-08-01
Fix for updated Bloomberg website

### LINKED ISSUE #324: Bloomberg Returning Parse Error
Hi, GnuCash user here. Just wanted to point out that the Bloomberg module has been failing to get online quotes and returning a parse error lately. It was working fine up until a few days or a week ago.

pghmcfc:
This should fix #324

The module must now use cookies to retrieve the quote data successfully. This is done using `HTTP::CookieJar::LWP`, which is a new (but small) dependency.

The module now provides the 'name' label in addition to the existing support for 'last', 'currency', 'symbol' and 'isodate'.

Conversion is also done from British pence to pounds where necessary. The associated test now includes a lookup that uses this feature.

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -154,7 +154,7 @@
 - module: Bloomberg.pm
   state: working
   added: 
-  changed: 2022-04-17
+  changed: 2023-08-01
   removed: 
   urls: https://www.bloomberg.com/quote/
   apikey: false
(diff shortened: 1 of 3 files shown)


bpschuck:
I was working on this myself yesterday. Pretty much the same code. I was also looking into trying to use JSON decode to parse the application/json that follows `__NEXT_DATA__` in the HTML. 
Thanks.

PR #340, opened 2023-08-15
New currencies in Stooq.pm

Kaligula0:
1) Currencies:
+EUR, HKD, HUF, JPY, USX
-ZAR (as unused)
2) Fix for commodities traded in:
- <span>$/</span>bbl, <span>$</span>/gal, <span>$</span>/t or <span>$</span>/mmBtu for energy commodities
- $/ozt, ¢/ozt or ¢/lb for metals
- ¢/bu or ¢/lb for grains
- $/t or ¢/lb for softs
- ¢/lb for fibers
- €/t for CO2 emission
and which don't have a currency A tag, thus no href attribute.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,8 +1,8 @@
 {{$NEXT}}
 	* Stooq.pm - Added new currencies and a fix for commodities' prices
 	* YahooWeb.pm - Skip rows in the price table where the prices ar "-".
-      This seems to happen sometimes with TIAA (and perhaps other) secyrutues
-      including TILIX and QCILIX
+	  This seems to happen sometimes with TIAA (and perhaps other) secyrutues
+	  including TILIX and QCILIX
 	* TSP.pm - Was not returning hash when the HTTP GET failed completely
 	  or the content did not contain the expected CSV file. - Issue #338
 	* BSEIndia.pm - Removed print when symbol not found - Issue #335
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Stooq.pm - Added new currencies and a fix for commodities' prices
 	* TSP.pm - Was not returning hash when the HTTP GET failed completely
 	  or the content did not contain the expected CSV file. - Issue #338
 	* BSEIndia.pm - Removed print when symbol not found - Issue #335
--- a/Changes
+++ b/Changes
@@ -1,6 +1,6 @@
 {{$NEXT}}
 	* Stooq.pm - Added new currencies and a fix for commodities' prices
-    * YahooWeb.pm - Skip rows in the price table where the prices ar "-".
+	* YahooWeb.pm - Skip rows in the price table where the prices ar "-".
       This seems to happen sometimes with TIAA (and perhaps other) secyrutues
       including TILIX and QCILIX
 	* TSP.pm - Was not returning hash when the HTTP GET failed completely
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -791,7 +791,7 @@
 - module: Stooq.pm
   state: working
   added: 2023-08-04
-  changed: ~
+  changed: 2023-08-15
   removed: ~
   urls:
     - https://stooq.com/q/?s=
--- a/lib/Finance/Quote/Stooq.pm
+++ b/lib/Finance/Quote/Stooq.pm
@@ -46,18 +46,25 @@ sub methods {
 our @labels = qw/symbol name open high low last bid ask date currency method/;
 
 my %currencies_by_link = (
+  '?i=21' => "EUR",
   '?i=23' => "GBP",
+  '?i=25' => "HKD",
+  '?i=30' => "HUF",
+  '?i=39' => "JPY",
   '?i=60' => "PLN",
   '?i=77' => "USD",
-  '?i=89' => "ZAR",
 );
 
 my %currencies_by_symbol = (
-  'p.' => "GBX",
-  '&pound;' => "GBP",
+  'p.'       => "GBX",
+  '&pound;'  => "GBP",
+  '&euro;'   => "EUR",
   'z\x{142}' => "PLN",
-  '\$' => "USD",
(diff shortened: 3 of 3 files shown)


bpschuck:
Please update Changes and Modules-README.yml. In Changes, most recent changes should be added immediately below `{{$NEXT}}` and above the previous change.
Thank you.

bpschuck:
@Kaligula0 - if you could please post some symbols to lookup that use those currencies?
Thank you.

Kaligula0:
> Please update Changes and Modules-README.yml. In Changes, most recent changes should be added immediately below `{{$NEXT}}` and above the previous change. Thank you.

OK, I will!

> @Kaligula0 - if you could please post some symbols to lookup that use those currencies? Thank you.

- EUR - [RAW.DE](https://stooq.com/q/?s=raw.de)
- HKD - [700.HK](https://stooq.com/q/?s=700.hk)
- HUF - [OTP.HU](https://stooq.com/q/?s=otp.hu)
- JPY - [7974.JP](https://stooq.com/q/?s=7974.jp)
- USX - used in commodities below
- energy commodities: <span>$/</span>bbl ([CB.F](https://stooq.com/q/?s=cb.f)), <span>$</span>/gal ([HO.F](https://stooq.com/q/?s=ho.f)), <span>$</span>/t ([LF.F](https://stooq.com/q/?s=lf.f)) or <span>$</span>/mmBtu ([NG.F](https://stooq.com/q/?s=ng.f))
- metals: $/ozt ([GC.F](https://stooq.com/q/?s=gc.f)), ¢/ozt ([SI.F](https://stooq.com/q/?s=si.f)) or ¢/lb ([HG.F](https://stooq.com/q/?s=hg.f))
- grains: ¢/bu ([ZC.F](https://stooq.com/q/?s=zc.f)) or ¢/lb ([ZL.F](https://stooq.com/q/?s=zl.f))
- softs: $/t ([CC.F](https://stooq.com/q/?s=cc.f)) or ¢/lb ([KC.F](https://stooq.com/q/?s=kc.f))
- fibers: ¢/lb ([CT.F](https://stooq.com/q/?s=ct.f))
- CO2 emission: €/t ([CK.F](https://stooq.com/q/?s=ck.f))

bpschuck:
Needs work. Currency returns `undef` for `GC.F`, `HG.F`, and others. 

```
bschuck@valhalla:~/tmp/finance-quote$ ./Examples/stockdump.pl stooq GC.F
Use of uninitialized value $currlink in hash element at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/Stooq.pm line 132.
Use of uninitialized value $currsymbol in hash element at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/Stooq.pm line 137.
Use of uninitialized value $currency in string eq at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/Stooq.pm line 162.
Use of uninitialized value $currency in string eq at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/Stooq.pm line 162.
$VAR1 = {
          'GC.Fsymbol' => 'GC.F',
          'GC.Fopen' => '1944.90',
          'GC.Fdate' => '08/25/2023',
          'GC.Fsuccess' => 1,
          'GC.Fcurrency' => undef,
          'GC.Fisodate' => '2023-08-25',
          'GC.Flow' => '1931.00',
          'GC.Fmethod' => 'stooq',
          'GC.Fhigh' => '1950.40',
          'GC.Fname' => 'Gold - COMEX (GC.F)',
          'GC.Flast' => '1939.90'
        };
```

Kaligula0:
Should be good now.

bpschuck:
Thanks again. While the module still works for commodities such as `GC.F`, may I suggest an edit to line 132. This will fix the following message printed to STDERR.

`Use of uninitialized value $currlink in hash element at /home/bschuck/tmp/finance-quote/lib/Finance/Quote/Stooq.pm line 132.`

Change `if ( $currencies_by_link{$currlink} ) {` to
`if ( ($currlink) && ($currencies_by_link{$currlink}) ) {`

And thanks for fixing the conflict caused by the most recent PR merge for YahooWeb.

Kaligula0:
Done :) Did the same in line 137.

bpschuck:
Which means that it is possible for `$currency` to not be set. This will cause issues everywhere else `$currency` is used for the remainder of the module. There are a few ways to account for that, I'm thinking the easiest is to change `if ( ($last) && ($date) ) {` to `if ( ($last) && ($date) && ($currency) ) {`. I'm on the fence about leaving `$currency` out of the returned hash if it cannot be parsed. This method would mean the subsequent lines referring to it would need to change so they are only executed if `$currency` is defined.

Kaligula0:
If it was only for GnuCash I'd choose the 1st solution, because GnuCash needs the currency. But I understand F::Q is for general purpose, therefore it has to be able to provide data for indexes, too. And that would mean choosing the 2nd solution.

bpschuck:
> I understand F::Q is for general purpose, therefore it has to be able to provide data for indexes, too.

No, not exactly. Besides, indexes from different exchanges would sort of have a currency. That being the currency native to the country where the exchange is. Also, to quote the opening line from `perldoc Finance::Quote`, "Finance::Quote - Get stock and mutual fund quotes from various exchanges". And the description, "This module gets stock quotes from various internet sources all over the world.". 

There is no hard requirement for F::Q to provide data for various indexes. This means the 1st option is okay as far as I'm concerned.

Thanks.

Kaligula0:
Now I see your point. If this is the case, then it's ok for me, too.
Should I change the line 149 from `if ( ($last) && ($date) ) {` to `if ( ($last) && ($date) && ($currency) ) {`?

bpschuck:
For now, yes, the quick easy one line fix. It can be fixed **if** users indicate they cannot get quotes for indexes.
Thanks.

Kaligula0:
Done 👍

PR #342, opened 2023-08-19
Skip rows in price table that don't have prices

mtalexander:
The YahooWeb interface sometimes returns tables that have rows with no prices.  Skip them and find the first row with prices. One security that shows this problem is TILIX.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,7 @@
 {{$NEXT}}
+        * YahooWeb.pm - Skip rows in the price table where the prices ar "-".
+          This seems to happen sometimes with TIAA (and perhaps other) secyrutues
+          including TILIX and QCILIX
 	* TSP.pm - Was not returning hash when the HTTP GET failed completely
 	  or the content did not contain the expected CSV file. - Issue #338
 	* BSEIndia.pm - Removed print when symbol not found - Issue #335
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -1109,7 +1109,7 @@
 - module: YahooWeb.pm
   state: working
   added: 2023-05-31
-  changed: 2023-07-16
+  changed: 2023-07-18
   removed:
   urls:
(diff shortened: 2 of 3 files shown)


bpschuck:
Please update Changes and Modules-README.yml. In Changes, most recent changes should be added immediately below `{{$NEXT}}` and above the previous change.
Thank you.

bpschuck:
@mtalexander,
Curious, when did this happen? I have not been able to reproduce.
Thanks.

mtalexander:
This was happening to various TIAA securities about a week ago.  I see that TILIX no longer shows the problem but QCILIX still does (although not on the most recent quote). See [QCILIX](https://finance.yahoo.com/quote/QCILIX/history?p=QCILIX).

bpschuck:
@mtalexander  Thanks again. I'm wondering if `volume` should not be returned if there is no value.

mtalexander:
I'm not sure I understand.  It seems that if there is no value there is no volume either, at least in the ones I've looked at.  Besides, GnuCash can't use the result if it has a volume and no price.  I must be missing something.

bpschuck:
Look at the data returned for the symbols you mentioned on the Yahoo web site. Open, high, low, and close will have values, but there is never a volume. I think that may be normal for all funds.

mtalexander:
Ooops, I misunderstood your comment.  Yes I agree that volume should not be returned if it has no value/

mtalexander:
I found two more bugs, one mine and one not.  I got confused (again) about Perl string vs. numeric comparisons. This caused it to return an errror if the first good line was between the third and tenth line.  Then the error message was lost because it was returning it with the wrong key.  Both are fixed in the last commits.

mtalexander:
This time I wanted a numeric comparison which is hard to do with a regex.

I'm going to be out of town and mostly out of EMail contact from Sunday through the 19th.  This probably doesn't matter unless someone finds something wrong with this after you merge it.

PR #343, opened 2023-08-22
Added YahooChart.pm module.

ka-patel:
Added YahooChart.pm module which retrieves historical data from Yahoo's chart API for last seven days and from that it provides the last good known data point as the pricing for the security.


bpschuck:
Likely not going to be implemented.

ka-patel:
OK. In that case please kindly reject the merge request of the code package.

PR #345, opened 2023-09-15
Correct a self-reference in TMX.pm documentation

dhgutteridge:


bpschuck:
@dhgutteridge,
Thank you for PR. We've been asking people to also update "Modules-README.yml" and "Changes" to minimize housekeeping done in preparation of pushing a release to CPAN. In "Changes" we have been adding the most recent to the top of the list under the `{{$NEXT}}` tag. In "Modules-README.yml", just update the `Changed: ` date for the module.

dhgutteridge:
Sorry about that, should be added as you need now. (These should be auto-squishable by GitHub, I think.)

bpschuck:
Not a problem. The "Hacker's Guide" notes those changes for new modules, but doesn't mention instructions or steps for those submitting PRs for fixes and updates. A ToDo for the module is creating a Contributing.pod that would contain more details than Hackers-Guide. When completed, the Contributing doc it would likely supersede the Hacker's Guide.
Thanks again.

PR #352, opened 2023-11-26
Fix scraper in Finanzpartner.pm

horas13:
The parsing of Finanzpartner page sometimes fails because the current parser only works if the current quote is less than the previous day's quote.  This fix checks for both cases and uses the correct value.

### DIFF
--- a/Changes
+++ b/Changes
@@ -8,6 +8,7 @@
 	* TSP.pm - Was not returning hash when the HTTP GET failed completely
 	  or the content did not contain the expected CSV file. - Issue #338
 	* BSEIndia.pm - Removed print when symbol not found - Issue #335
+	* Finanzpartner.pm - Fix scraper, did not work if quote was higher than the previous day's quote.
 
 1.58      2023-08-12 10:59:05-07:00 America/Los_Angeles
 	* Consorsbank.pm - New module - PR #329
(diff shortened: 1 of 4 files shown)


bpschuck:
@horas13,
Thank you for the contribution. It would be appreciated if you can also update the `Changes` and `Modules-README.yml` files. This helps with final housekeeping we do before pushing a new release to CPAN. In the `Changes` file, most recent changes are added to the top of list under `{{$NEXT}}`. 
Another helpful step is checking that the all the tests in `t/finanzpartner.t` are still valid. Update or remove those symbols for securities that are no longer traded.
Thank you.

horas13:
Ok, I updated `Changes`.
`Modules-README.yml` does not have much info (yet), so there is nothing to change. 

The test passes, but probably the coverage was quite poor to begin with, I will see if I can improve it a bit to cover at least this fix.

bpschuck:
On 11/27/23 06:15, horas13 wrote:

> Ok, I updated |Changes|.
> |Modules-README.yml| does not have much info (yet), so there is nothing 
> to change.

Well, state can be changed to working,changed to when you this PR, urls 
to contain the URL used to fetch data, testfile the name of the test 
script in the "t/" directory, and testcases a list of symbols used in 
the testfile.

horas13:
I extended the testfile and updated `Modules-README.yml`.
The tests currently fail without my patch and work with the patch. That might change however, as the behaviour depends on the history of the quotes. Of course using multiple and highly anti-correlated stocks might help, but I am to lazy for that. :-)

PR #362, opened 2024-03-18
Fix Bloomberg module, now extracting data from JSON blob embedded in the HTML

### LINKED ISSUE #360: Bloomberg Parse Error
Hello, 

The Bloomberg module has been returning parse errors lately. From what I found, there are two issues: One has to do with the encoding/decoding of the HTML page, and the other is a parsing error.

I have found that removing line 42 in the module solves the encoding/decoding issue: `'Accept-Encoding' => 'br',`

As for the parsing issue, changing line 58 to the following solves it: 
`my $date  = $tree->look_down(_tag=>'span', 'class'=>qr/^marketLastUpdate_exchangeDelay___PZEn/)->right();`

pghmcfc:
This should fix #360.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Bloomberg.pm - Changed module to extract data from JSON structure embedded within the HTML - Issue #360
 	* NSEIndia.pm - Eliminated need to use temp folders by storing file data from URL into a variable.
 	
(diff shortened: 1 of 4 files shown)


bpschuck:
@pghmcfc  I know I'm picky, but prefer to have most recent changes at top of `Changes` file. So the line for this change would be above the line for `NSEIndia.pm`. Thanks.

PR #371, opened 2024-04-13
Update the forked source.

ka-patel:


ka-patel:
Not necessary.

bpschuck:
As I've stated some time ago. The there is no need for the YahooChart module.

PR #385, opened 2024-04-23
New module: BorsaItaliana.pm - getting quotes from Borsa Italiana (Italian traded bonds) using ISIN

fabrica64:
New module to get Italian traded bonds quotes from the Borsa Italiana (Milan Stock Exchange, Euronext group) official website.


bpschuck:
FYI - The module name gets added in 2 places in `Quote.pm`. Also add it to the POD section at the bottom please.

We also prefer that people `rebase` out of date forks/branches instead of merging. Makes for a cleaner commit tree.

Thanks for the contribution.

fabrica64:
> FYI - The module name gets added in 2 places in `Quote.pm`. Also add it to the POD section at the bottom please.

Added

> We also prefer that people `rebase` out of date forks/branches instead of merging. Makes for a cleaner commit tree.

Noted, thanks!

PR #387, opened 2024-04-25
Updated yahooJSON.pm to handle EU consent in robust way.

ka-patel:
Updated yahooJSON.pm to handle EU consent in robust way which dovetails into getting proper cookies and corresponding crumb to auth correctly.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 {{$NEXT}}
 	* BorsaItaliana.pm - New module for Borsa Italiana, Italian traded bonds using ISIN
 	* YahooWeb.pm - Issue #377. Modified YahooWeb to account for changes from Yahoo.
+	* Modified yahooJSON.pm module in order handle EU consent redirects better.
 
 1.61      2024-04-18 21:34:24-07:00 America/Los_Angeles
 	* SIX.pm - Changed lookup for currency, added lookups for symbol and last. Issue #380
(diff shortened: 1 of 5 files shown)


ka-patel:
This version is tested via VPN in Germany and the US; it tested as working.

This removes hitting undocumented https://fc.yahoo.com host for initial cookies.

ka-patel:
Noted.

bpschuck (changes requested):
@ka-patel 
As previously mentioned, since this has not been merged, please move the line added to `Changes` above the `BorsaItaliana.pm` line.
While testing, running the test file `t/yahoojson.t` failed on a few things. I realized that this file has probably not been updated when changes to the module were rushed last year. It would be appreciated if you look at the failures and make the appropriate adjustments. An example is that original module would return "Sourced from Yahoo Finance (as JSON)" as the value for `exchange`. Since that can now be a variety of exchanges, that check can probably be removed. I have not looked at the other failures in detail yet.
Thanks again.
Bruce

bpschuck:
I fixed `t/yahoojson.t`.

bpschuck:
The YahooJSON currency module is using the v8 API, which so far doesn't require special handling.

Run tests locally. There's a few reasons it's not done in the CI/CD.

bpschuck:
@ka-patel,

I could open a discussion thread if you would like to have a conversation about the CI/CD and testing.

Thanks.

Bruce

ka-patel:
I don't think that is not necessary. I was interested in personally understanding the testing framework employed here. Your prior response makes it clear why is $ONLINE_TEST not enabled. If you are looking for effort to make the test cases bit more robust than I try to help wherever I can.

PR #399, opened 2024-05-14
Added my own modules named GenericExecutor and YahooChart.

ka-patel:
Added my own modules named GenericExecutor and YahooChart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,32 @@
+# NOTE: 
+
+This repo includes updates necessary to add YahooChart and GenericExecutor 
+module, which are personally developed and tested. 
+
+YahooChart employs multi-threading to fetch data when available and the
+module does not need to ascertain whether day is a trading day, holiday 
+or anything else that may not be a trading day. Currently it downloads 
+last seven days of quote data set and then returns last valid data set as 
+the valid pricing. The return pricing is always the adjusted pricing that 
+accounts for splits and dividends.
+
+This module can also retrieve currency but this seems to be very limited 
+as to what Yahoo's Chart API will return.
+
+
+GenericExecutor module is technically a bridge between GNUCash and OS via
+Finance::Quote. This module will invoke an executor at the operating system 
+level, such as python, ruby, go, or any other interpreter, along with a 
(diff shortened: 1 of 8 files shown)


bpschuck:
@ka-patel 

Please be aware F::Q is akin to an API that given a source (aka method name) and list of symbols will return the hash containing the single data values for each symbol. Or the exchange rate for two currencies. Modules that return anything else can not and will not be incorporated into the official Finance::Quote package and namespace. 

If you are developing a specialized module for personal use you may want to look into using the FQ_LOAD_QUOTELET environment variable. 

From the Quote.pm POD:

> With no arguments, new creates a Finance::Quote object with the default methods. If the environment variable FQ_LOAD_QUOTELET is set, then the contents of FQ_LOAD_QUOTELET (split on whitespace) will be used as the argument list. This allows users to load their own custom modules without having to change existing code. If any method names are passed to new or the flag '-defaults' is included in the argument list, then FQ_LOAD_QUOTELET is ignored.

> When new() is passed one or more class name arguments, an object is created with only the specified modules loaded. If the first argument is '-defaults', then the default modules will be loaded first, followed by any other specified modules. Note that the FQ_LOAD_QUOTELET environment variable must begin with '-defaults' if you wish the default modules to be loaded.

ka-patel:
Bruce, indeed F::Q is akin to an API given that GnuCash and other apps use it underneath it -- the contract is fixed and not likely to change without much debate. By mistake the commit intended for my fork ended up to official F::Q repo. 

I am familiar with FQ_LOAD_QUOTELET settings and I believe my personally developed ones are adhering to that. If you see, both modules return hash in following manner:

```
    return wantarray() ? %info : \%info;
    return \%info;
```
and populate info using:

```
    $info{ $commodity, 'xxx' } = 'yyy';
```
hence, they should be following the spirit of the official framework and the contract. I don't mind if you look one over at sources of two modules and advise if it is otherwise.

Pointer to POD is appreciated on FQ_LOAD_QUOTELET.

ka-patel:
I don't mind if you remove this commit, branch and pull request out from official F::Q history. If you want to keep, then that is fine as well.

PR #422, opened 2024-09-01
FinanceAPI.pm - New Module

bpschuck:
FinanceAPI.pm is new module. See https://financeapi.net/. An API Key is required. The URL used by the `financeapi` method is https://yfapi.net/v6/finance/quote?. 
Data is available from US and other exchanges. The free basic API Key allows 100 queries per day rate limited to 300 per minute.
It provides for the multiple/failover methods "nyse", "nasdaq", and "usa".

### DIFF
--- a/lib/Finance/Quote/FinanceAPI.pm
+++ b/lib/Finance/Quote/FinanceAPI.pm
@@ -148,6 +148,8 @@ sub financeapi {
 
       # Check for stocks traded in pence instead of pounds
       # Convert GBp or GBX to GBP (divide price by 100).
+      # GBP.L - pence
+      # GBPG.L - pounds
 
       if ( ($info{$stock,"currency"} eq "GBp") ||
          ($info{$stock,"currency"} eq "GBX")) {
--- a/lib/Finance/Quote/FinanceAPI.pm
+++ b/lib/Finance/Quote/FinanceAPI.pm
@@ -1,0 +1,236 @@
+#!/usr/bin/perl -w
+# vi: set ts=2 sw=2 noai ic showmode showmatch:  
+#
+#    Copyright (C) 2024, Bruce Schuck <bschuck@asgard-systems.com>
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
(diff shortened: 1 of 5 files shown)


lib/Finance/Quote/FinanceAPI.pm
```diff
@@ -0,0 +1,276 @@
...
+package Finance::Quote::FinanceAPI;
+
+use strict;
+use warnings;
+
+use Encode qw(decode);
+use HTTP::Request::Common;
+use JSON qw( decode_json );
+
+use constant DEBUG => $ENV{DEBUG};
```
ka-patel, line 33: Only recommendation here is to change it to:

`use constant DEBUG => $ENV{FQ_FINANCEAPI_DEBUG} || $ENV{FQ_DEBUG} || $ENV{DEBUG};`

Reason is that there are too many things out there that use DEBUG env var for all sort of things generically which obsefucates debug messages. This will allow to tune for specific emitter as much specific as needed to troubleshoot.

lib/Finance/Quote/FinanceAPI.pm
```diff
@@ -0,0 +1,276 @@
...
+
+      $name     = $quote->{'quoteResponse'}{'result'}[0]{'longName'};
+      $open     = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketOpen'};
+      $high     = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketDayHigh'};
+      $low      = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketDayLow'};
+      $last     = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketPrice'};
+      $volume   = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketVolume'};
+      #$currency = $quote->{'quoteResponse'}{'result'}[0]{'financialCurrency'};
+      $currency = $quote->{'quoteResponse'}{'result'}[0]{'currency'};
+      $date     = $quote->{'quoteResponse'}{'result'}[0]{'regularMarketTime'};
```
ka-patel, line 133: I'd declare an intermediate var for `$quote->{'quoteResponse'}{'result'}[0]` and then use that for clarity. That is just about coding style ...

ka-patel (commented):
See my comments.

PR #444, opened 2024-10-26
Update MorninstarJP to use xml data from www.wealthadvisor.co.jp

### LINKED ISSUE #443: www.wealthadvisor.co.jp change
www.wealthadvisor.co.jp, which is used by `MorningstarJP.pm` (ref. #261) has changed.

It looks like the fund quote is available as non-public API in the form of `https://www.wealthadvisor.co.jp/xml/`__FUNDCODE__`.xml`, though it retrieves the whole history of quote values and becomes several hundreds kilo bytes size.

pkryger:
I took a liberty to rewrite the module. I wasn't sure how to interpet the data that is returned in the xml, please double check it. I decided to use the fund name in `name` field instead of its symbol. Please let me know if anything needs to be updated.

Please also see the note about Shift_JIS encoding. I am no expert in this matter and not really sure how to verify if I picked the right one. I guess a funds with names that contain a few conflicting character could be chosen (Yen sign?) and verify that it has been decoded correctly. See this https://github.com/steve-m-hay/XML-Encoding/blob/e48f2c7/maps/Japanese_Encodings.msg. Alternatively, the `XML::Encoding` module could be updated, or `Finance::Quote` could be shipped with it's own encmap file. Unfortunately, I can't read kanji, which makes these changes that more difficult.

This fixes #443

### DIFF
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -805,9 +805,9 @@
 - module: MorningstarJP.pm
   state: working
   added: TBD
-  changed: 2023-04-08
+  changed: 2024-10-26
   removed: ~
-  urls: https://www.wealthadvisor.co.jp/FundData/DownloadStdYmd.do?fnc=
+  urls: https://www.wealthadvisor.co.jp/xml/
   apikey: false
   notes: ~
   testfile: morningstarJP.t
--- a/dist.ini
+++ b/dist.ini
@@ -76,6 +76,8 @@ Module::Load = 0.36
 Mozilla::CA = 0
 Test2 = 1.302167
 URI::Escape = 3.31
+XML::Parser = 0
+XML::Encoding = 0
 
 [Prereqs / TestRequires]
 Date::Range = 0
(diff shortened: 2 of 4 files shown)


bpschuck:
Thanks, but I have my own rewrite.

bpschuck:
Using XML::LibXML instead of XML::Parser. The original module returned the symbol as `name`, so using the KISS principle, not going to bother with dealing with encodings. Since the original returned a value for `nav` and not return `price` or `last`, continuing to return `nav` along with `price` and/or `last`.

bpschuck:
@pkryger 

FYI for future contributions. Never assume every function is successful. 

    my $parser = XML::Parser->new(Handlers => {Start => \&$on_start},
                                        ProtocolEncoding => 'x-sjis-unicode');

    $parser->parse($xml);
    ### Extracted: $name, $code, $date->ymd, $price, $nav
          
Checking that the parse completed as expected would be desired. We want to avoid the module exiting abnormally due to attempting to utilize undefined values.

pkryger:
@bpschuck thanks for the pointers. Will keep that in mind for next time.

PR #455, opened 2025-01-09
Return actual multipliers for CurrencyFreaks.pm module

ka-patel:
Updated CurrencyFreaks.pm to return actual multipliers for base currency rather then computed conversion rate.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Updated CurrencyFreaks.pm to return actual multipliers for base currency rather then computed conversion rate
 	* Added methodinfo hash to Fool.pm
 	* Fixed MorningstarJP.pm - Issue #443 - PR #451
 
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -355,12 +355,12 @@
 - module: CurrencyRates/CurrencyFreaks.pm
   state: working
   Added: 2024-05-09
-  changed: ~
+  changed: 2025-01-09
   removed: ~
   urls: https://api.currencyfreaks.com/v2.0/rates/latest?apikey=YOUR_APIKEY&symbols={CURRENCYCODE},{CURRENCYCODE},...
   apikey: true
   notes: |
-    Newly module as of 5/9/2024
+    New module as of 5/9/2024
   testfile: currency.t
   testcases:
     - USD EUR
--- a/lib/Finance/Quote/CurrencyRates/CurrencyFreaks.pm
+++ b/lib/Finance/Quote/CurrencyRates/CurrencyFreaks.pm
@@ -91,7 +91,7 @@ sub multipliers
 
   my $json_data = decode_json ($body);
 
-  if ( !$json_data->{'rates'}->{$from} || !$json_data->{'rates'}->{$to} ) {
+  if ( !$json_data->{'rates'}->{${from}} || !$json_data->{'rates'}->{${to}} ) {
     return;
   }
 
--- a/lib/Finance/Quote/CurrencyRates/CurrencyFreaks.pm
+++ b/lib/Finance/Quote/CurrencyRates/CurrencyFreaks.pm
@@ -118,7 +118,8 @@ sub multipliers
     return ($b, $a);
   }
 
-  return (1.0, $rate);
+  # return actual multipliers that are relative to a third currency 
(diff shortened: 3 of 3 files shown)


bpschuck (changes requested):
The `module: CurrencyRates/CurrencyFreaks.pm` block in Modules-README.yml needs to be updated (`changed:`).

I've decided that going forward, environment variables take precedence over arguments that may have been passed in. At the same time, if no API key is found the method should simply return nothing.

```
# CurrencyFreaks is permitted to use an environment variable for API key 
# (for backwards compatibility).
# New modules should use the API_KEY from args.

$this->{API_KEY} = $ENV{'CURRENCYFREAKS_API_KEY'};
$this->{API_KEY} = $args->{API_KEY} if (ref $args eq 'HASH') and (exists $args->{API_KEY});

return $this;
```

should be rewritten as

```
# CurrencyFreaks is permitted to use an environment variable for API key 
# (for backwards compatibility).
# New modules can use the API_KEY from args.

$this->{API_KEY} =
  $args->{API_KEY} if (ref $args eq 'HASH') and (exists $args->{API_KEY});
$this->{API_KEY} =
  $ENV{'CURRENCYFREAKS_API_KEY'} if exists $ENV{'CURRENCYFREAKS_API_KEY'};

# Return nothing if API_KEY not set
return unless ($this->{API_KEY});

return $this;
```

ka-patel:
I am for “I've decided that going forward, environment variables take precedence over arguments that may have been passed in.” given numerous API key are used but I failed to understand “At the same time, if no API key is found the method should simply return nothing.”. Normally good practice is to permeate some sort of message upwards so that the end-user eventually gets a proper indication of the semantic cause but I suspect something more compelling reason(s) causing to take this tactical.

 

I’ll update it as requested.

bpschuck:
> I am for “I've decided that going forward, environment variables take precedence over arguments that may have been passed in.” given numerous API key are used but I failed to understand “At the same time, if no API key is found the method should simply return nothing.”. Normally good practice is to permeate some sort of message upwards so that the end-user eventually gets a proper indication of the semantic cause but I suspect something more compelling reason(s) causing to take this tactical. 

None of the other currency rate modules return an error message, so I'm just going with the past flow. I'm not sure if GnuCash looks for an error message in the data returned. There is no automatic failover for the currency rates either.

But yes, at some point it may be worth opening a dialogue with the GnuCash devs and having cleaner fails for currency lookups.

Thanks.

bpschuck:
Huh?! Changes should have been in the same branch as your original pull request...

You also may want to look at the rest of Modules-README.yml. `changed:` is just a date...

ka-patel:
In line with thinking of API keys, proposing change from:

 

use constant DEBUG => $ENV{DEBUG}

 

to:

 

use constant DEBUG => ($ENV{DEBUG} || $ENV{FQ_DEBUG} || $ENV{FQ_CURRENCYRATES_DEBUG} ||$ENV{FQ_CURRENCYFREAKS_DEBUG})

 

thoughts?

bpschuck:
NO

bpschuck:
I've also done some testing, and I do not see any issues with the module in its current state.

```
bschuck@valhalla:~$ currency-lookup.pl NZD JPY
NZD->JPY = 87.62
bschuck@valhalla:~$ FQ_CURRENCY=YahooJSON currency-lookup.pl NZD JPY
NZD->JPY = 87.62
bschuck@valhalla:~$ FQ_CURRENCY=CurrencyFreaks currency-lookup.pl NZD JPY
NZD->JPY = 87.6262598202082
```

And I am a big believer in "if it's not broken, do not fix it!". 

Unless you can demonstrate to my satisfaction that there is an issue, I'm going to close this PR.

ka-patel:
Sorry for the delay in response – was bit pre-occupied with other excitements going on so couldn’t address your emails quicker.

 

Anyways, closing it is fine. It is NOT broken by any means.

 

After studying the hackers guide for currency rate modules thoroughly, I wanted to bring it in line with same spirit as stated, as in:

 

Currency rate modules should avoid carrying out arithmetic and leave it to

Finance::Quote to use the multipliers to compute the final conversion value.

 

Which this exchange rate provider can provide hence the update. May be I am reading more than meets the eye.

PR #466, opened 2025-02-02
 Improve AEX (exchange, close, time)

moenny:
This PR improves the AEX module:

- add label `exchange`, `close`, `time`

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,6 @@
 {{$NEXT}}
-	* Added label exchange, close, time to AEX.pm - PR #465
+	* Added methodinfo hash to AEX.pm - PR #466
+	* Added label exchange, close, time to AEX.pm - PR #466
 	* Fixed XETRA.pm - Issue #460
 	* Added methodinfo hash to Fool.pm
 	* Fixed MorningstarJP.pm - Issue #443 - PR #451
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Added label exchange, close, time to AEX.pm - PR #465
 	* Fixed XETRA.pm - Issue #460
 	* Added methodinfo hash to Fool.pm
 	* Fixed MorningstarJP.pm - Issue #443 - PR #451
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -5,7 +5,7 @@
 - module: AEX.pm
   state: working
   added: TBD
-  changed: 2024-05-11
+  changed: 2025-02-02
   removed: ~
   urls:
     - https://live.euronext.com/en/search_instruments/
@@ -21,6 +21,8 @@
     - AMG
     - LVMH
     - NL0000009165
+    - BE0974293251
+    - IE0002MXIF34
 #
 - module: AIAHK.pm
   state: removed
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -5,7 +5,7 @@
 - module: AEX.pm
   state: working
   added: TBD
-  changed: 2025-02-02
+  changed: 2025-02-05
   removed: ~
   urls:
     - https://live.euronext.com/en/search_instruments/
--- a/lib/Finance/Quote/AEX.pm
+++ b/lib/Finance/Quote/AEX.pm
@@ -24,24 +24,35 @@ use warnings;
 use constant DEBUG => $ENV{DEBUG};
 use if DEBUG, 'Smart::Comments';
 
-use vars qw( $AEX_URL);
-
 use LWP::UserAgent;
 use Web::Scraper;
 use String::Util qw(trim);
 
 # VERSION
 
-sub methods { 
-  return (dutch => \&aex,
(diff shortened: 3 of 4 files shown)


bpschuck:
Thanks.

Changes have not been incorporated into the hacker's guide, but please refer to this section from AlphaVantage.pm. As modules are updated or introduced, the way methods and labels are defined and presented is being changed.

```
our $DISPLAY    = 'AlphaVantage';
our $FEATURES   = {'API_KEY' => 'registered user API key'};
our @LABELS     = qw/date isodate open high low close volume last net p_change/;
our $METHODHASH = {subroutine => \&alphavantage, 
                   display => $DISPLAY, 
                   labels => \@LABELS,
                   features => $FEATURES};

sub labels {
  my %m = methodinfo(); return map {$_ => [@{$m{$_}{labels}}] } keys %m;
}

sub methodinfo {
    return ( 
        alphavantage => $METHODHASH,
        canada       => $METHODHASH,
        usa          => $METHODHASH,
        nyse         => $METHODHASH,
        nasdaq       => $METHODHASH,
    );
}

sub methods {
  my %m = methodinfo(); return map {$_ => $m{$_}{subroutine} } keys %m;
}
```

We've also been asking contributors to modify the `Changes` and `Modules-README.yml` file to help reduce housekeeping needed before pushing releases to CPAN. Note that most recent changes are at the top of the `Changes` file. For the `Modules-README.yml` file, symbols used for testing are noted, and date of last update to module indicated on the `changed:` field (YYYY-MM-DD format).

Also note that since GnuCash is the primary consumer of F::Q derived data, adding labels for quote data not used by GnuCash is a low priority.

Thank you.

pkryger:
> I think the module should be renamed from `AEX.pm` to `Euronext.pm`.

If this is something that goes through, please consider updating GnuCash as well, such that it can display the new module name.

bpschuck:
You appear to be assuming that GnuCash and Finance::Quote are 2 pieces of the same project. They are not. You may want to start a discussion on the gnucash-devel email list about the GnuCash-Finance::Quote interface.

pkryger:
> Functionally there would be no incompatibility.

That's correct. To a degree. Just for posterity here's what I have taken out from last time I worked on synchronising GnuCash and Finance::Quote.

GnuCash is capable of handling "unknown" sources, that is a source that is reported by Finance::Quote or a source that is found in a GnuCash user file, but the source doesn't have corresponding entry hardcoded into GnuCash application. The former is usually a caused by a new source name reported by Finance::Quote. The latter is a source name that used to be a valid name reported by Finance::Quote, and which was subsequently removed from Finance::Quote. As you can see human readable (perhaps "user friendly" is a better term?) names are a part of GnuCash code. What's more such human readable names are subjected to translation in GnuCash. However, each source addition to Finance::Quote will cause it to appear in "unknown sources" combo box in GnuCash. And each source removal will cause a breakage for GnuCash users that happened to use that particular name, IIRC the source will still appear as a known one but will be disabled (greyed out) in GnuCash interface. (For that one could devise a n automatic migration mechanism, but that's another discussion). As a bonus there's a GnuCash Wiki with Finance::Quote sources (including some sources that don't exist any more) that is meant to help GnuCash users configure their quote sources. 

Probably the best option is what @bpschuck has suggested.

bpschuck:
> I am aware that finance-quote and GnuCash are independent projects.
> 
> Several finance-quote modules (e.g. AEX.pm, Tradegate.pm) should be renamed. Therefore I suggest not to rename them in this PR. I think the renaming of the affected modules would be better in a dedicated PR together with a corresponding PR at GnuCash.

Keeping F::Q and GC in sync is akin to playing a game of Whack-A-Mole. Now if you are volunteering to take the reigns of F::Q...

bpschuck:
> That's why I don't want to rename module by module.
> I am still looking at the F::C modules. Once I have an overview, I might make a PR for the renaming of the affected F::C modules and a PR for GC.

I wouldn't waste your time. I'm not going merge a PR that renames and moves around a large number of module and method names. Yeah, it is a bit random and messy, but I'm a huge fan of "if it isn't broken, don't fix it".

And don't forget, not everyone uses the current GnuCash release. Many Linux users are happy with the GnuCash release available from the OS repository for their distro. For example, my main desktop at home is running Ubuntu 22.04 (I'm been lazy on updating to 24.04) with GnuCash 4.8 installed from Canonical. Obviously because various website changes necessitate keeping F::Q mostly current, it is not the canonical package. 

> I'm not sure what you mean by this. But I feel very comfortable in the role of a contributor. ;-)

By "taking the reigns" I mean the person who does the triage on issues and PRs, prepares and pushes releases (and pre-releases) to CPAN, keeps the documentation and hackers guide up to date, keeps the info on the SourceForge site up to date, etc. It's not too bad when things are slow, but when a heavily used module like YahooJSON stops working it could get hectic.

Like I said earlier on, you may want to open a dialogue on the gnucash-devel email list and especially John R.

PR #476, opened 2025-03-07
Adding Hong Kong stock support to GoogleWeb

stevel408:
With the recent and repeated problems with YahooJson, I think the GoogleWeb option is becoming more and more viable and valuable. Here's a very small tweak for it to handle stock traded in the Hong Kong Stock Exchange (HKEX), with ticket symbols such as "2828", "0700", etc.

The logic today is to fetch from a Google Finance web URL for a ticker such as "TSLA", and then find an "A" anchor in the format of something like "TSLA:NASDAQ" and then fetch that page. There is a limited set of "exchanges" that are supported there, and I just added "HKG". I'm sure there are others (like Toronto TSX, London LON, etc.), if the reviewer(s) are willing, please feel free to add those too.

Thank you for the good work!

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* GoogleWeb - Add support for Hong Kong Exchange - PR #476
 	* YahooJSON - Adjust other currency fields when data is returned in GBp, ZAc, or ILA.
 	* Added label wkn, ìsin, close, ask, bid, p_change, time to OnVista.pm
 	* Modified OnVista.pm to enable search by ISIN and WKN and search for ETFs
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -642,7 +642,7 @@
 - module: GoogleWeb.pm
   state: working
   added: 2023-09-16
-  changed: 2023-12-16
+  changed: 2025-03-06
   removed: ~
   urls: https://www.google.com/finance/
   apikey: false
--- a/lib/Finance/Quote/GoogleWeb.pm
+++ b/lib/Finance/Quote/GoogleWeb.pm
@@ -84,8 +84,8 @@ sub googleweb {
       $tree = HTML::TreeBuilder->new;
       if ($tree->parse_content($body)) {
         #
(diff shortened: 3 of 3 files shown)


bpschuck:
> With the recent and repeated problems with YahooJson, I think the GoogleWeb option is becoming more and more viable and valuable. Here's a very small tweak for it to handle stock traded in the Hong Kong Stock Exchange (HKEX), with ticket symbols such as "2828", "0700", etc.

> The logic today is to fetch from a Google Finance web URL for a ticker such as "TSLA", and then find an "A" anchor in the format of something like "TSLA:NASDAQ" and then fetch that page. There is a limited set of "exchanges" that are supported there, and I just added "HKG". I'm sure there are others (like Toronto TSX, London LON, etc.), if the reviewer(s) are willing, please feel free to add those too.

Thank you for your contribution.

The GoogleWeb was a quick addition at a time that we were seeing frequent changes affecting the YahooJSON module. Which is why it was written to support just the typical US exchanges.

The issue with other exchanges for GoogleWeb is that some of the logic would need to be rewritten to assist in differentiating which exchange a user is looking for the price from. Take NWG (NatWest Group plc) for example. With no suffix, it exists as an NYSE security and a London exchange (LON) security. Depending on the order in the HTML returned from the first URL fetch, the TreeBuilder `look_down` call will return the first one. So the code would need to be changed to parse a suffix from the symbol, and maybe map some so the symbol "NWG.L" which is valid for YahooWeb, YahooJSON, AlphaVantage, and others, might perhaps get mapped to "NWG:LON" and the second URL fetch may not be needed. At the same time, allow the ":LON" and other suffixes so users used to getting data from the Google finance pages can use a familiar symbol.

Good things for a future roadmap. But for now, I'm okay with the module as is (but I will merge this PR into the master sometime soon).

Thank you.

stevel408:
Thank you and really appreciate it. Let me know when/if you decide to tackle this down the road, I'm happy to pitch in to help, at least with testing, etc.

PR #478, opened 2025-03-13
Comdirect: add etf, more labels and EXCHANGE feature

moenny:
improves the Comdirect module:
- fix ETF search support
- add label `time`, `p_change`, `ask`, `bid`, `wkn`, `exchange` 
- a new label `exchanges` contains the list of available marketplaces
- the marketplace can be selected using the 'EXCHANGE' feature. This feature is already under development for other modules.

Update 2025-03-16:
- fix encoding
  - i.e.: CH0038863350 name: Nestl**é**

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Added ETFs search, more labels and EXCHANGE feature to Comdirect.pm - PR #478
 	* YahooJSON - Adjust other currency fields when data is returned in GBp, ZAc, or ILA.
 	* Added label wkn, ìsin, close, ask, bid, p_change, time to OnVista.pm
 	* Modified OnVista.pm to enable search by ISIN and WKN and search for ETFs
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -238,7 +238,7 @@
 - module: Comdirect.pm
   state: working
   added: TBD
-  changed: 2024-10-18
+  changed: 2025-03-13
   removed: ~
   urls: https://www.comdirect.de/inf/search/all.html?SEARCH_VALUE=
   apikey: false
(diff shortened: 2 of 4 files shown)


bpschuck:
I'm thinking that returning a list of exchanges is a bit much. It is something GnuCash will ignore it and could be considered beyond the scope of F::Q's primary purpose. Same for "time". F::Q isn't meant for real time lookups, but was originally written for once-a-day updates of security prices for GnuCash.

bpschuck:
And to eliminate the "Parsing of undecoded UTF-8 will give garbage when decoding entities..." warning, perhaps change line 79 from

`unless ($tree->parse(encode_utf8($body))) {`

to

`unless ($tree->parse($body)) {`

bpschuck:
> The list of stock exchanges is useful for selecting the name on the respective website. Background: The list of stock exchanges depends on the security and the name of the stock exchange sometimes differs on the various websites.

> I am using F::Q for a not yet published project and would like to continue using F::Q and also contribute improvements to F::Q. I think that the contributions can also be useful for other secondary purposes.

Modules for F::Q can be written and utilized without being part of the published package (and this repository). See the F::Q man page where is discusses using the `FQ_LOAD_QUOTELET` environment variable. Again, probably 99% of F::Q users are using it from GnuCash to update security prices daily.

bpschuck:
I need to think about this more. I think F::Q v1.65 will be released soon without your last 4 PRs. As I said, the prime function of F::Q is the stock price utility for GnuCash. While it's awesome that you and others have found it useful for other purposes, care needs to be taken so additional features do not affect how it interfaces with GnuCash.
One thing that we would like to discourage are F::Q based tools that are used for many lookups and/or multiple times a day. The sites that data is scraped from at some point will notice that in their logs and take measures to block or otherwise hinder access to quote data. Recently at least one source switched to using Incapsula's Imperva service, and that module had to be removed from F::Q. Including, configuring, and utilizing a JavaScript engine to circumvent that protection is something that is beyond the scope of the vast majority of GnuCash users.

PR #465, opened 2025-02-02
 Improve Tradegate (ISIN, ask, bid, time)

moenny:
This PR improves the Tradegate module:

- fix ETF support
- add label `isin`, `open`, `ask`, `bid`, `time`

PR #481, opened 2025-03-16
Sinvestor: add feature 'EXCHANGE', some fixes

moenny:
a bigger cleanup:
- changes from #465 (Tradegate.pm) adopted
- like #478:
  - a new label exchanges contains the list of available marketplaces
  - the marketplace can be selected using the 'EXCHANGE' feature.
- ~~Tradegate.pm now uses Sinvestor.pm via EXCHANGE feature~~
- ~~XETRA.pm now uses Sinvestor.pm via EXCHANGE feature~~

- fix encoding and don't store '-' as value
  - i.e.: CH0038863350 name: Nestl**é** 
- make the code more readable and reliable (searches for strings instead of blindly accessing columns)
- fix some currencies (i.e. "Schweizer Franken" -> "CHF")
- add label `ask`, `bid`, `source`
- fix and update tests

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,5 @@
 {{$NEXT}}
 	* Added EXCHANGE feature to Sinvestor.pm and some fixes - PR #481
-	* Tradegate now uses Sinvestor.pm via EXCHANGE feature - PR #481
-	* XETRA now uses Sinvestor.pm via EXCHANGE feature - PR #481
 	* Added EXCHANGE feature to OnVista.pm - PR #480
 	* Added EXCHANGE feature to Consorsbank.pm and some fixes - PR #479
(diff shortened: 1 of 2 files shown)


bpschuck:
Please do not reference variables, functions, or methods from other modules. This can cause a maintenance nightmare if a module no longer works and is removed while other modules have a dependency on it.

bpschuck:
In addition, Modules-README.yml must be updated to reflect changes to symbols used in the test scripts.

bpschuck:
Again, the primary application for Finance::Quote is GnuCash. It does not have the ability to pass warnings back to the user. Nor does it currently have the ability to pass options/arguments to F::Q. Each Finance::Quote::FooBar module must be self-contained for lack of a better word, aside from using the few public methods from the main Quote.pm module.

PR #517, opened 2025-07-30
Fix MorningstarCH.pm - basically copying MorningstarUK.pm.

fritteli:


bpschuck:
Thank you once again. Now I have to decide. Since there already is redundancy with XETRA.pm and you submitted this fix for MorningstarCH, is adding SwissFundData.pm really necessary. Especially from a maintenance point view.
Decisions, decisions, decisions...

fritteli:
You're right 😄 
I leave it up to you (and I'm glad I don't have to decide). If you decide against merging SwissFundData, that's no problem.

PR #525, opened 2025-08-28
Fix TesouroDireto after site update

grmontesino:
Get Tesouro Direto quotes from the "rendimento-resgatar" csv. The json file previosly used as source isn't available anymore after an update to the site's layout.

Patch mostly based on Deka.pm module.


grmontesino:
> Thank you for your contribution. If you could please modify "Changes" and "Modules-README.yml" appropriately, that would be appreciated. Thank you.

Docs updated; thanks.

PR #544, opened 2025-11-24
Add more fields to YahooJSON module

chromatic:
These additional fields primarily affect NYSE- and NASDAQ-listed stocks.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,6 @@
 {{$NEXT}}
 	* YahooJSON.pm - Add additional fields/labels - PR #544
+	* New Module, USBonds.pm
 	* MarketWatch.pm - Added headers to fix 401 code - Issue #542
 	* Modified AEX.pm to return no data available is currency indicator is '%' (bonds return percentage of par/facevalue) - Issue #539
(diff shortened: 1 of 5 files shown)


bpschuck:
I am curious, how many lookups per day are you doing? The majority of users, mostly GnuCash users, query once a day for end-of-day prices. Typical GnuCash users may have somewhere between 1 and 2 dozen securities to look up. 
I'm concerned that it is the outlier users that make many queries a day that will cause the financial sites to find ways to make it difficult for scripting ways to fetch data.

BTW, people can create their own private modules/methods and utilize the `FQ_LOAD_QUOTELET` environment variable to load them. Granted, if this PR is not merged, you will need to fix your own module if Yahoo once again adds hoops to jump through. If you browse through the CPAN module listing, you will find a handful of modules that other developers have created and made available to the public.

bpschuck:
> I also query once a day but the portfolio is larger than 20 securities.

Less than 100? If you are just using one source, and have no need for the others F::Q offers, why not use a source such as FinanceAPI and utilize their APIs directly? Using FinanceAPI or one of the other token based sites means that there is a far less chance of any changes that would affect your ability to fetch data.

Issue #542, opened 2025-11-13
marketwatch.pm - unauthorized message


omoknen:
FYI to other users.  Currently marketwatch.pm no longer works.  It has an unauthorized error code.

error:
Finance::Quote reported a failure for symbol VWIAX: Error retrieving quote for VWIAX. Attempt to fetch the URL https://www.marketwatch.com/investing/stock/VWIAX resulted in HTTP response 401 (Unauthorized)

bpschuck:
Thank you for the report. Unfortunately this will likely mean the module will be removed and not fixed.

omoknen:
Thanks for the quick reply.  All of you have done a lot of work over the years, and it is appreciated.  It is best to keep focus on the modules that can be maintained.  Removing that module is probably the best answer.

We need only basic "last" quotes for our needs.

We now have a primary (yahoojson) and two backups (googleweb, yahooweb).
[Hope this helps anyone else trying to find simple replacements]

Once again Thanks

bpschuck:
> Thanks for the quick reply. All of you have done a lot of work over the years, and it is appreciated. It is best to keep focus on the modules that can be maintained. Removing that module is probably the best answer.

There is just so many rounds of "Whak-A-Mole" one has the patience for. 

> We need only basic "last" quotes for our needs.

I'm curious. Do you use Finance::Quote from within GnuCash, KMyMoney, by itself, or as part of a tool you developed yourself?

> We now have a primary (yahoojson) and two backups (googleweb, yahooweb). [Hope this helps anyone else trying to find simple replacements]

For US stocks and Funds there is also "fool" (Motley Fool) as a non-apikey method/module. 

We do have the modules listed on the Wiki - https://github.com/finance-quote/finance-quote/wiki/Documentation#individual-sources

Depending on how many securities you retrieve pricing for, I strongly suggest one or more of the methods requiring an API Key.

> Once again Thanks

You are most welcome. Thank you for being understanding that removing the broken module can be the solution.

bpschuck:
@ka-patel 

Thanks, but don't go too deep. I have found that when using a browser console, aside from disabling cache in the cosole, all cookies need to be cleared before fetching a quote from https://www.marketwatch.com/investing/stock/ to force the initial GET to result in a 301 status code to capture the mechanism that does the authentication.

> They are resorting to complex authentication schemes which does some way require result of execution of java scripts to make sure that impression shown is falling on eyeballs.

[...]

PR #561, opened 2026-04-26
Add CurrencyRates/UniRate.pm

rob-browncc:
## Summary

Adds `Finance::Quote::CurrencyRates::UniRate` — a new pluggable currency-rate source backed by [UniRate](https://unirateapi.com). UniRate covers 1000+ currencies (fiat and crypto) with daily updates and a generous free tier.

```perl
$q = Finance::Quote->new(currency_rates => {order   => ['UniRate'],
                                            unirate => {API_KEY => ...}});

$value = $q->currency('18.99 EUR', 'USD');
```

## Why another currency source?

- **Broad crypto + fiat coverage** — covers symbols some existing sources don't (1000+ codes).
- **Free tier is enough for current rates** — this module only uses `/api/rates`, which is on the free plan; no pay-gate behaviour to surprise downstream users.
- **Pluggable failover slot** — slots cleanly into the existing `order =>` chain alongside ECB / Fixer / OpenExchange / etc., so users can layer it as primary or fallback.

## What's in this PR

- `lib/Finance/Quote/CurrencyRates/UniRate.pm` — new module (~90 LOC, mirrors the `CurrencyFreaks` / `AlphaVantage` pattern: `parameters()`, `new()`, `multipliers()`, with the small-rate inverse-flip).
- Wires `UniRate` into `@CURRENCY_RATES_MODULES` and `SEE ALSO` in `lib/Finance/Quote.pm`.
- New `Modules-README.yml` entry (state: working, apikey: true, testfile: `currency.t`, four test cases).
- `Changes` entry under `{{$NEXT}}`.
- `t/currency.t` subtest gated on `TEST_UNIRATE_API_KEY`, mirroring the layout used for the other API-key-backed sources.

## How it was tested

`prove -Ilib t/currency.t` with `ONLINE_TEST=1 TEST_UNIRATE_API_KEY=...`:

```
# Subtest: UniRate
    1..10
    ok 1
    ok 2 - 100.00 USD -> EUR = 85.4883990932263
    ok 4 - 1.00 GBP -> IDR = 23286.8656060185
    ok 6 - 1.23 IDR -> CAD = 9.75069016599033e-05
    ok 8 - 10.00 AUD -> AUD = 10.00
    ok 9 - identity check
    ok 10 - 20.12 ZZZ -> GBP failed as expected
ok 7 - UniRate
```

Cross-checked live rates against the YahooJSON subtest in the same run — values agree to within ~0.2% (independent sources, expected drift).

POD also passes the author test: `TEST_AUTHOR=1 prove -Ilib t/01-pod.t` → `Files=1, Tests=56, Result: PASS`.

## Implementation notes

- Single `GET /api/rates?from=…&to=…&api_key=…` per call. Response is `{"rate": "0.8548…"}`.
- Sets `Accept: application/json` and a `User-Agent: finance-quote-unirate/0.01` header on the request — the upstream sits behind Cloudflare, which 403s the default `libwww-perl/X.YZ` UA. (Same workaround pattern as the headers added to `MarketWatch.pm` for issue #542.)
- API key is read from the `unirate` hash inside `currency_rates` only — no env-var fallback, consistent with the comment in `CurrencyFreaks.pm` that env-var fallback is reserved for the legacy modules.
- Small-rate inverse path matches the `CurrencyFreaks` / `AlphaVantage` shape (`if ($rate < 0.001) { ... }`).

Happy to fold in any review feedback — naming, error handling, additional test pairs, etc.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,4 +1,5 @@
 {{$NEXT}}
+	* Added CurrencyRates/UniRate.pm.
 
 1.69      2026-04-18 11:47:37-07:00 America/Los_Angeles
 	* Sinvestor.pm - trim whitespace before parsing date and price
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -476,6 +476,23 @@
     - GBP IDR
     - IDR CAD
 #
+- module: CurrencyRates/UniRate.pm
+  state: working
+  Added: 2026-04-26
+  changed: ~
+  removed: ~
+  urls: https://api.unirateapi.com/api/rates?from={CURRENCYCODE}&to={CURRENCYCODE}&api_key=YOUR_APIKEY
+  apikey: true
+  notes: |
+    Free tier covers /api/rates (1000+ currencies, fiat + crypto).
+    Running test file requires TEST_UNIRATE_API_KEY environment variable.
+  testfile: currency.t
+  testcases:
+    - USD EUR
+    - GBP IDR
+    - IDR CAD
+    - AUD AUD
+#
 - module: CurrencyRates/YahooJSON.pm
   state: working
   Added: 2023-04-09
--- a/lib/Finance/Quote.pm
+++ b/lib/Finance/Quote.pm
@@ -58,6 +58,7 @@ use vars qw/@ISA @EXPORT @EXPORT_OK @EXPORT_TAGS
     Fixer
     OpenExchange
     TwelveData
+    UniRate
     YahooJSON
 /;
 
@@ -1706,6 +1707,7 @@ http://www.gnucash.org/
   Finance::Quote::CurrencyRates::Fixer,
   Finance::Quote::CurrencyRates::OpenExchange,
   Finance::Quote::CurrencyRates::TwelveData,
+  Finance::Quote::CurrencyRates::UniRate,
   Finance::Quote::CurrencyRates::YahooJSON,
   Finance::Quote::AEX,
   Finance::Quote::ASEGR,
--- a/lib/Finance/Quote/CurrencyRates/UniRate.pm
+++ b/lib/Finance/Quote/CurrencyRates/UniRate.pm
@@ -44,6 +44,10 @@ sub new
 
   ### UniRate->new args : $args
 
+  # UniRate supports an environment variable for the API key for parity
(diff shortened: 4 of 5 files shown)


bpschuck:
@rob-browncc,
Thank you for your contribution. I will try to review this PR during the next week (4/27/26 -> 5/1/26). It will get merged if I don't find any issues or have suggestions.
A big thank you for updating "housekeeping" files such as Changes and Modules-README.

bpschuck:
@rob-browncc 

Please add the ability to get the API Key from the environment variable `UNIRATEAPI_API_KEY`. The Twelvedata and Financeapi currency rate modules also have this ability.

Thank you.

rob-browncc:
@bpschuck thanks for the review! Done in 33c8b7c — `UNIRATEAPI_API_KEY` env var with the explicit `API_KEY` in the `unirate` hash still winning, matching the Twelvedata and FinanceAPI patterns. POD section updated to document both paths.

bpschuck:
> @bpschuck thanks for the review! Done in [33c8b7c](https://github.com/finance-quote/finance-quote/commit/33c8b7c8bf55b81cdf951632bfd35ae5a33f7e88) — `UNIRATEAPI_API_KEY` env var with the explicit `API_KEY` in the `unirate` hash still winning, matching the Twelvedata and FinanceAPI patterns. POD section updated to document both paths.

Not a problem, thanks for the quick response.
I've been on the fence about which method should take precedence. I'm a fan of being able to test different keys -- perhaps a premium versus a basic -- without having to change code. For this, the environment variable would need take precedence. The original contributor who added the ability to pass in the key in the hash had the opinion that using environment variables should be eliminated. You may even find comments in some of the modules noting that.

That said, I hope to do some of my own sanity checks for this PR soon and maybe merge it before the end of the weekend.

bpschuck:
The PR has been merged. 

Would it be possible for you to update https://wiki.gnucash.org/wiki/Online_Quotes#Finance::Quote_Sources_-_Currency_Sources? 

At some point I need to start on the Currency section of https://github.com/finance-quote/finance-quote/wiki/Documentation#currency.

Thank you.

PR #568, opened 2026-06-13
Fix tab indentation in Modules-README.yml (Union.pm entry)

vincentl:
### Problem
`Modules-README.yml` does not parse as YAML. The `Union.pm` entry uses **tab characters** to indent its `methods:` key and the content of its `notes:` block. YAML forbids tabs for indentation, so any YAML loader fails.  

### Fix
Replace the tabs with spaces (2-space for the key, 4-space for the `notes` block content), matching the convention used by every other entry in the file. No data changed.


bpschuck:
Good evening sir.
If this fix is also part of one of the other PRs you submitted, this PR can be closed with no action.
Thanks.

vincentl:
I keep the PRs separate. This one and #571  are simple edits, the other 2 PRs are for specific sources.

PR #570, opened 2026-06-13
Add new source - Finnhub.io

vincentl:
### Summary
Adds a Finance::Quote module for [finnhub.io](https://finnhub.io), which
offers a **free, personal-use** API key (no credit card, 60 calls/min) for
US equities, ETFs, forex, and crypto.

### How it works
- **`/quote`** (required) → `last`, `open`, `high`, `low`, `close`
  (previous close), `net`, `p_change`, and the date from the trade
  timestamp. Works for equities and ETFs.
- **`/stock/profile2`** (best-effort) → `currency` and `name`. Listed
  companies only; gracefully degrades for ETFs/other instruments, where
  currency defaults to USD (consistent with the free tier's US focus).
- Handles `401` (invalid key), `429` (rate limit), and the all-zero
  response Finnhub returns for unknown symbols.

### API key
Read from the `finnhub => {API_KEY => ...}` constructor hash or the
`FINNHUB_API_KEY` environment variable — same pattern as `AlphaVantage.pm`.
**No key is stored in the repo.** The test reads `TEST_FINNHUB_API_KEY` and
skips when it (or `ONLINE_TEST`) is unset.

### Methods provided
`finnhub`, `usa`, `nyse`, `nasdaq`.

### Checklist (per CONTRIBUTING.md)
- [x] Module with POD: `lib/Finance/Quote/Finnhub.pm`
- [x] `# VERSION` comment after the `use` statements
- [x] Registered in `@MODULES` and SEE ALSO in `Quote.pm` (alphabetical)
- [x] Test with passing + failing symbols: `t/finnhub.t`
- [x] Entry added to `Changes` and `Modules-README.yml`
- [x] `t/01-pod.t`, `t/04-critic.t`, `podchecker` pass

### Verification (system perl + local::lib)
- Online test through the framework: **20/20 pass** (AAPL, MSFT, SPY; BOGUS fails as expected)
- Method registration and labels confirmed via `Finance::Quote->new('Finnhub')`

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,5 @@
 {{$NEXT}}
-	* New module, Finnhub.pm. Fetches quotes from https://finnhub.io using a
-	  free personal-use API key (FINNHUB_API_KEY). Provides the finnhub, usa,
-	  nyse, and nasdaq methods for US equities, ETFs, forex, and crypto.
+	* New module, Finnhub.pm. Fetches quotes from https://finnhub.io using a free personal-use API key.
 
 	* Tradegate.pm: remove explicit "our $VERSION" assignment that collided
 	  with the dzil-injected version, producing a duplicate declaration and
--- a/Changes
+++ b/Changes
@@ -1,4 +1,8 @@
 {{$NEXT}}
+	* New module, Finnhub.pm. Fetches quotes from https://finnhub.io using a
+	  free personal-use API key (FINNHUB_API_KEY). Provides the finnhub, usa,
+	  nyse, and nasdaq methods for US equities, ETFs, forex, and crypto.
+
 	* Tradegate.pm: remove explicit "our $VERSION" assignment that collided
 	  with the dzil-injected version, producing a duplicate declaration and
 	  an "our variable $VERSION redeclared" warning. Use a bare # VERSION
--- a/Modules-README.yml
+++ b/Modules-README.yml
@@ -676,6 +676,30 @@
    - "LU0856992614"
    - "LU1720050803"
 #
+- module: Finnhub.pm
+  state: working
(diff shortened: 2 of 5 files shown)


vincentl:
Rebased on `master` to resolve conflicts.

bpschuck:
Vincent,
If you could also please create a GnuCash PR and modify `libgnucash/engine/gnc-commodity.cpp`, and the GnuCash Wiki that has information about the online sources.
Thanks.

bpschuck:
@vincentl 
Perhaps a simple one second sleep is warranted between each URL fetch? Or maybe first check the number of symbols to be fetched. If less than 60, don't sleep between fetches. If more than 60, sleep between each one.

vincentl:
Good point on respecting the API rate. I'll push an update. It looks like you beat me to the wiki update.  I'll do the gnucash pull request.

bpschuck:
> Good point on respecting the API rate. I'll push an update. It looks like you beat me to the wiki update. I'll do the gnucash pull request.

Yeah, I updated both the GnuCash Wiki (https://wiki.gnucash.org/wiki/Online_Quotes#Finance::Quote_Sources) and the one here (https://github.com/finance-quote/finance-quote/wiki/Documentation#financequote-sources).

Thanks again.

vincentl:
Just pushed a change that follows how the pacing is done in AlphaVantage. It also drops the original plan of 2 API calls per symbol. The intent was the second call got the company profile and currency - but the free tier is only for US securities, so we can assume USD and only execute 1 API call per symbol.

vincentl:
Opened PR Gnucash/gnucash#2247

Issue #573, opened 2026-06-22
License text includes out of date FSF address


pghmcfc:
The `Documentation/License` and `LICENSE` files include this text:
```
 Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 Everyone is permitted to copy and distribute verbatim copies
 of this license document, but changing it is not allowed.
```
and, later on:
```
    You should have received a copy of the GNU General Public License along
    with this program; if not, write to the Free Software Foundation, Inc.,
    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```
This address has been out of date since August 2024 (https://www.fsf.org/blogs/community/fsf-office-closing-party). The FSF now suggests (https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html) this text instead:
```
 Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
 <https://fsf.org/>
 Everyone is permitted to copy and distribute verbatim copies
 of this license document, but changing it is not allowed.
```
and
```
You should have received a copy of the GNU General Public License
along with this program; if not, see
<https://www.gnu.org/licenses/>.
```
The header comments of most of the modules also refer to the same address.

bpschuck:
@pghmcfc,

Thank you for pointing this out. But looking at the license files of the Perl module sources that currently exist under my ~/.cpan/build directory, **many** of them contain the same outdated FSF address. As this minor "housekeeping" issue doesn't seem to a problem for the rest of the Perl community, I am closing this issue. Although it is something to keep in mind in the future when addressing other issues and bug fixes.

PR #574, opened 2026-06-24
Modernize and reconcile project licensing

vincentl:
Although issue #573 was closed, I have found AI tools are quite good at monotonous tasks like normalizing code. I tried Claude with Opus 4.8 and it noticed that several modules were released under GPLv3. Since the other modules that carried a notice said GPLv2 **or later**, Claude pointed out the project is effectively GPLv3 (because GPLv3 is not compatible with GPLv2) and suggested moving FQ officially to GPLv3.

If there is interest in polishing this PR, happy to make changes. If now is not a good time to consider this update, just close the PR.

Here is a summary of the 56 module files on the master branch:

## Finance::Quote license inventory (56 modules, mutually exclusive)

| Primary code-header grant | Before | After |
|---------------------------|-------:|------:|
| GPLv2-or-later            |     51 |    50 |
| GPLv3-or-later            |      3 |     6 |
| No license header         |      2 |     0 |
| **Total**                 | **56** | **56** |

What moved between the columns:
- `Finance/Quote.pm`: GPLv2-or-later → GPLv3-or-later (project umbrella)
- `Deka.pm`, `TreasuryDirect.pm`: no header → GPLv3-or-later (headers added)
- All other 53 modules: grant unchanged

Subset note (not a separate row): 4 of the GPLv2-or-later modules — FTfunds,
MorningstarCH, MorningstarUK, SwissFundData — additionally offer themselves
"under the same terms as Perl itself" in a POD section. That dual grant was
left untouched.

## Notice-text modernization (independent of grant)

| License-notice style                       | Before | After |
|--------------------------------------------|-------:|------:|
| Old FSF postal address (51 Franklin St)    |     50 |     0 |
| Old FSF postal address (59 Temple Place)   |      1 |     0 |
| `http://…/licenses/` pointer               |      3 |     0 |
| No pointer (no header)                      |      2 |     0 |
| `https://www.gnu.org/licenses/` pointer    |      0 |    56 |
| **Total**                                   | **56** | **56** |

## Change Details

Boilerplate modernization (all modules under lib/):
  - Replaced the obsolete FSF postal address paragraph ("write to the Free Software Foundation, Inc., 51 Franklin Street ..." and, in USBonds.pm, the even older "59 Temple Place" address) with the current FSF form: "if not, see <https://www.gnu.org/licenses/>." This matches the wording the FSF now recommends in "How to Apply These Terms".
  - Normalized the three GPLv3 modules' pointer from http to https.
  - Each module keeps its original version grant (v2-or-later stays v2-or-later, v3-or-later stays v3-or-later, "same terms as Perl" is untouched) and all copyright attribution lines are preserved verbatim.

Missing headers:
  - Added a license/copyright header to the two modules that had none, Deka.pm and TreasuryDirect.pm, at the project default (GPL v3 or later).

Project-level license:
  - Added a top-level LICENSE file (full GPLv3 text), generated with Software::License::GPL_3 so it matches what dzil now produces. The repo previously had no LICENSE; the full text existed only in the release tarball.
  - dist.ini: license = GPL_2 -> GPL_3. This is the single source of truth; LICENSE, META.json and META.yml are regenerated from it, so they cannot drift from the declared license.
  - Documentation/License: collapsed the duplicate GPLv2 text into a pointer to the canonical root LICENSE, with a note on the per-module grants.

Documentation:
  - lib/Finance/Quote.pm POD (and the generated README.md) COPYRIGHT & LICENSE section: updated to GPL v3 or later, added the https pointer, and added a paragraph explaining the per-module grants and why the aggregate is GPLv3-or-later. Existing copyright lines are unchanged.
  - CONTRIBUTING.md: added a "Licensing and the file header" section with the canonical header template for new modules and a note never to downgrade an existing author's grant.

Regression guard:
  - Added t/06-license-headers.t (author test, gated by TEST_AUTHOR) that fails if any module lacks a license notice, reintroduces the obsolete postal address, or uses an http (non-https) licenses URL. Run: TEST_AUTHOR=1 prove -Ilib t/06-license-headers.t


bpschuck:
Not implementing at this time.

bpschuck:
In poking around I've come across many Perl modules that simply just have

"This is free software, you may use it and distribute it under the same terms as
Perl itself."

In both the License file and embedded in POD or comments. Aka, the KISS principle.

vincentl:
For a fresh project, this would be simple policy. For FQ, you can't change the license of existing modules without getting each author's consent. Given the history of the project, it seems infeasible to track down authors.

PR #589, opened 2026-08-31
Added NPSNAV.pm module to fetch NAV for India's National Pension Scheme (NPS) funds.

ka-patel:
Added NPSNAV.pm module to fetch NAV for India's National Pension Scheme (NPS) funds.

PR #590, opened 2026-09-04
TreasuryDirect.pm: repair fetching, which failed for every CUSIP

allen-ziegenfus:
`TreasuryDirect.pm` returns `Parse error` for every CUSIP against the current
site. Behind that are two pricing bugs that the fetch failure currently hides.

## Reproducing

`912810QT8` is one of the two CUSIPs in `t/treasurydirect.t`, and is a live
security in the FedInvest file:

```sh
perl -MFinance::Quote -e '
  my %i = Finance::Quote->new()->fetch("treasurydirect", "912810QT8");
  printf("success=%s errormsg=%s\n",
         $i{"912810QT8","success"} // 0, $i{"912810QT8","errormsg"} // "(none)");'
```

```
success=0 errormsg=Parse error
```

## Why the fetch fails

Three causes, each sufficient on its own:

1. **The POST runs through `wget`.** Many systems do not have it, and its
   absence is not checked, so the captured output is empty, no table is found,
   and every symbol reports a parse failure. Now uses the Finance::Quote user
   agent, as the rest of the distribution does.
2. **The form no longer takes `priceDate.month` / `priceDate.day` /
   `priceDate.year`.**
3. **The endpoint is Spring-backed.** It rejects a POST carrying no `_csrf`
   token bound to the session cookie set when the form is fetched (403), and
   answers a successful POST with a redirect, which LWP does not follow for
   POST unless asked.

Installing `wget` does not help — causes 2 and 3 still apply, and the request
the module makes by hand returns `HTTP/1.1 403`.

## Two pricing bugs

Both return a plausible number rather than an error, so neither shows up as a
failure.

**Averaging against an unquoted side.** `price` was the mean of BUY and SELL
unconditionally. FedInvest publishes `BUY` as `0.000000` for securities quoted
on one side only, so the mean is about half the real price. Running unmodified
1.71 against the 2026-09-02 file prices **430 securities correctly and 33 at
roughly half**. The mean is now taken only when both sides are quoted, falling
back to the end of day price or the quoted side.

**`bid` and `ask` were inverted.** FedInvest's `BUY` is what an investor pays,
so it is the ask; `SELL` is what an investor receives, so it is the bid.
`BuyPrice` is at or above `SellPrice` in every record — 389 of 463 strictly
above on 2026-09-02, the remainder equal — so the previous mapping reported a
negative spread on every security. `price` is a mean and is unaffected.

Since the fetch failures above stop the module reaching this code, both were
confirmed by driving unmodified 1.71 against a page captured from the live
site, with a stub `wget` on `PATH` so the released parsing and pricing code
runs exactly as shipped:

```
912810QT8   success=1  price=78.87   bid=78.890625   ask=78.843750
912797RS8   success=1  price=49.99   bid=0.000000    ask=99.989972
91282CHY0   success=1  price=50.02   bid=0.000000    ask=100.031250
```

`912810QT8` is quoted on both sides and prices correctly, which is what makes
this easy to miss: not a broken module, a wrong number for a subset of
securities.

A container that reproduces all of the above from a clean build is at
<https://github.com/allen-ziegenfus/fq-treasurydirect-repro>.

## Reading the XML rather than the HTML table

The same form offers the day's prices as XML at `securityPriceDetail` with
`fileType=xml`, against a published schema (`FedInvestPrice_v2_0_0.xsd`), with
named fields — `Cusip`, `BuyPrice`, `SellPrice`, `EODPrice` — instead of a
table whose columns have to be counted.

`XML::LibXML` is already required by `ECB.pm` and `MorningstarJP.pm`, so this
adds no dependency, and `HTML::TableExtract` is no longer needed here (it stays
pinned for the five other modules that use it).

To be clear about what this does and does not fix: the table has gained an
`END OF DAY` column since the module was written, but it was appended after
the fields being read, so the positional map is still correct. This removes the
assumption rather than a present-day fault.

## Also in this change

- **Dropped `ssl_opts(verify_hostname => 0)`**, which disabled TLS hostname
  verification on the shared user agent for every module fetched afterwards in
  the same `Finance::Quote` object.
- **The retry loop now reports.** Exhausting all four days returned an empty
  list with no `success` and no `errormsg`, contrary to §3.9 of the Hackers
  Guide. A form yielding no CSRF token likewise now says so, instead of
  surfacing the resulting 403 as a network fault.
- **Timeout is bounded** only when the caller has set none. The module
  previously overwrote `timeout` on the shared agent; with that removed the
  default was LWP's 180s across up to eight requests.
- **Date handling** goes through `store_date`. 1.71 regex-matched `Prices For:`
  with a hand-rolled month table and assigned `date`/`isodate` directly, which
  §3.8 asks modules not to do.
- **`keys(%bonds) > 1`** treated a file containing exactly one security as
  empty.

## Changes visible to callers

- `bid` and `ask` swap meaning, per the inversion above.
- `price` is returned at the six decimal places FedInvest publishes, where it
  was previously rounded to two — a format change for anyone parsing or
  displaying it.
- Labels go from
  `method source symbol rate bid ask price date isodate`
  to
  `method source symbol rate maturity bid ask eod price date isodate currency`.
- Labels are **omitted rather than returned as zero** when the file did not
  quote them, so `eod` is absent during the trading day, and the unquoted side
  is absent for a one-sided security.

## Testing

- `t/treasurydirect.t` — 28 tests with `ONLINE_TEST=1`. Adds assertions that
  the price is not a halved average and that the ask is not below the bid, the
  two regressions above.
- `t/treasurydirect-parse.t` — **new, 18 tests, offline.** Drives the parse and
  the price selection from canned documents covering a two-sided note, a
  one-sided bill, a completed day with an end of day price, a security with
  nothing quoted, an empty file for a weekend, input that will not parse, and a
  comment node inside a `<Security>`. The parse and price selection were
  factored into `parse_prices` and `price_from` to make this possible.
- `00-store-date.t`, `01-pod.t`, `02-pod-coverage.t`, `05-data-dumper.t` pass;
  `04-critic.t` reports no violations at severity 5. `03-kwalitee.t` fails
  `has_buildtool`, `has_manifest` and `has_meta_yml`, identically on a pristine
  `master` checkout, since dzil generates those at build time.
- Verified across all 463 securities priced on 2026-09-02 that the XML read
  returns the same `BuyPrice` and `SellPrice` the HTML table does.


allen-ziegenfus:
Unrelated to this change, but I ran into it while rebasing onto current
`master` and it seems worth flagging: **`Modules-README.yml` no longer parses.**

`248f18a` (#589) indents four lines with tabs, and YAML does not permit tabs
for indentation:

```
966      Module to fetch the latest Net Asset Values (NAVs) of India's
967 		   National Pension Scheme (NPS) funds.
968    testfile: npsnav.t
969    testcases:
970 		   - SM001001
971 		   - SM001002
972 		   - SM008001
```

```sh
perl -MYAML -e 'YAML::LoadFile("Modules-README.yml")'
```

```
YAML Error: Invalid element in sequence
   Code: YAML_LOAD_ERR_BAD_SEQ_ELEMENT
   Line: 967
```

The v1.71 tag loads cleanly at 94 entries, so this is new. `YAML::PP` v0.41
(the parser `dist.ini` lists) fails at the same line.

Replacing those four tab indents with spaces is all it takes. I have left it
out of this PR since it belongs to a different change, but I am happy to send
it separately if that is useful.

While there: `YAML::PP` also reports duplicate `notes:` keys in three entries,
which predates #589 and is a separate matter.

```
Sinvestor.pm  lines 1078 and 1081
Tradegate.pm  lines 1230 and 1233
XETRA.pm      lines 1486 and 1489
```

The last value wins, so the earlier `notes:` in each is silently discarded.

bpschuck:
Thank you for discovering and fixing this.

However, in "Changes", it is preferred that the most recent changes are at the top. So this change should be above the NPSNAV addition. Also, only a brief description of the change is necessary. No need to describe the changes in detail.

Remove treasurydirect-parse.t. In my opinion it would just be another file to change when treasury direct makes changes that affect the content of the output.

As for the recent addition of NPSNAV to Modules-README.yml, please feel free to make the correction.

allen-ziegenfus:
Thanks for the quick look. All three done, as separate commits on top so they
are easy to read against the original.

- `Changes`: moved above the NPSNAV entry and cut from ten entries to four.
  Kept the two that change behaviour for anyone already using the module — the
  bid/ask meaning, and price going from two decimals to six — since those seem
  worth a user noticing. Happy to trim further.
- `t/treasurydirect-parse.t`: removed. I left `parse_prices` and `price_from`
  factored out, as they read more clearly than the same work inline, but
  nothing depends on them being separate.
- `Modules-README.yml`: the four tab-indented lines in the NPSNAV entry are now
  spaces, and the file loads again — 95 entries.

One thing I did not touch: `YAML::PP` also rejects the file for duplicate
`notes:` keys in `Sinvestor.pm`, `Tradegate.pm` and `XETRA.pm`. That predates
NPSNAV and `YAML.pm` tolerates it, so I left it alone rather than widen this
PR. Say the word if you would like it in here or as a separate change.

bpschuck:
>     * `Changes`: moved above the NPSNAV entry and cut from ten entries to four.
>       Kept the two that change behaviour for anyone already using the module — the
>       bid/ask meaning, and price going from two decimals to six — since those seem
>       worth a user noticing. Happy to trim further.

Please do. Note that historically, there is only one line per Module. If others want details, they can either look at the PR or commits to glean more specifics.

> One thing I did not touch: `YAML::PP` also rejects the file for duplicate `notes:` keys in `Sinvestor.pm`, `Tradegate.pm` and `XETRA.pm`. That predates NPSNAV and `YAML.pm` tolerates it, so I left it alone rather than widen this PR. Say the word if you would like it in here or as a separate change.

Yeah, I'll have to fix these. Since the Modules-README.yml file was a "work in progress" added not too long ago, and it's not parsed for use elsewhere, syntax checking as you can see was not a priority. If it was, we (well, when there was a "we"), may have added a `0X-parse-modules-readme.t` to be executed if `TEST_AUTHOR` was set. Although, that may be a good idea anyway...

That said, thank you again. Much appreciated when someone sees and corrects issues in what I call a game of "Whac-A-Mole" (https://en.wikipedia.org/wiki/Whac-A-Mole).

bpschuck:
Running into something bizarre.

Running `dzil build` or `dzil test` on your is not executing `[Git::NextVersion]`, and not replacing `# VERSION` in the modules with `our $VERSION = '1.72'; # VERSION` as expected.

Example:

If I clone from my repo, I get:

```
bschuck@valhalla:~/tmp/finance-quote$ dzil build
[DZ] beginning to build Finance-Quote
[Git::NextVersion] Bumping version from 1.71 to 1.72
[DZ] guessing dist's main_module is lib/Finance/Quote.pm
[@Basic/ExtraTests] rewriting author test xt/author/pod-syntax.t
[DZ] writing Finance-Quote in Finance-Quote-1.72
[MarkdownInRoot] overriding README.md in root
[DZ] building archive with Archive::Tar::Wrapper
[DZ] writing archive to Finance-Quote-1.72.tar.gz
[DZ] built in Finance-Quote-1.72
```

But cloning from your repo:

```
bschuck@valhalla:~/tmp$ git clone git@github.com:allen-ziegenfus/finance-quote.git
Cloning into 'finance-quote'...
remote: Enumerating objects: 11847, done.
remote: Counting objects: 100% (117/117), done.
remote: Compressing objects: 100% (68/68), done.
remote: Total 11847 (delta 64), reused 78 (delta 43), pack-reused 11730 (from 2)
Receiving objects: 100% (11847/11847), 2.85 MiB | 8.38 MiB/s, done.
Resolving deltas: 100% (6695/6695), done.
bschuck@valhalla:~/tmp$ cd finance-quote/
bschuck@valhalla:~/tmp/finance-quote$ git switch -c treasurydirect-fix 
Switched to a new branch 'treasurydirect-fix'
bschuck@valhalla:~/tmp/finance-quote$ dzil build
[DZ] beginning to build Finance-Quote
[DZ] guessing dist's main_module is lib/Finance/Quote.pm
[@Basic/ExtraTests] rewriting author test xt/author/pod-syntax.t
[DZ] writing Finance-Quote in Finance-Quote-1.21
[MarkdownInRoot] overriding README.md in root
[DZ] building archive with Archive::Tar::Wrapper
[DZ] writing archive to Finance-Quote-1.21.tar.gz
[DZ] built in Finance-Quote-1.21
```

bpschuck:
> Running into something bizarre.
> 
> Running `dzil build` or `dzil test` on your is not executing `[Git::NextVersion]`, and not replacing `# VERSION` in the modules with `our $VERSION = '1.72'; # VERSION` as expected.
> 
> Example:
> 
> If I clone from my repo, I get:
> 
> ```
> bschuck@valhalla:~/tmp/finance-quote$ dzil build
> [DZ] beginning to build Finance-Quote
> [Git::NextVersion] Bumping version from 1.71 to 1.72
> [DZ] guessing dist's main_module is lib/Finance/Quote.pm
> [@Basic/ExtraTests] rewriting author test xt/author/pod-syntax.t
> [DZ] writing Finance-Quote in Finance-Quote-1.72
> [MarkdownInRoot] overriding README.md in root
> [DZ] building archive with Archive::Tar::Wrapper
> [DZ] writing archive to Finance-Quote-1.72.tar.gz
> [DZ] built in Finance-Quote-1.72
> ```
> 
> But cloning from your repo:
> 
> ```
> bschuck@valhalla:~/tmp$ git clone git@github.com:allen-ziegenfus/finance-quote.git
> Cloning into 'finance-quote'...
> remote: Enumerating objects: 11847, done.
> remote: Counting objects: 100% (117/117), done.
> remote: Compressing objects: 100% (68/68), done.
> remote: Total 11847 (delta 64), reused 78 (delta 43), pack-reused 11730 (from 2)
> Receiving objects: 100% (11847/11847), 2.85 MiB | 8.38 MiB/s, done.
> Resolving deltas: 100% (6695/6695), done.
> bschuck@valhalla:~/tmp$ cd finance-quote/
> bschuck@valhalla:~/tmp/finance-quote$ git switch -c treasurydirect-fix 
> Switched to a new branch 'treasurydirect-fix'
> bschuck@valhalla:~/tmp/finance-quote$ dzil build
> [DZ] beginning to build Finance-Quote
> [DZ] guessing dist's main_module is lib/Finance/Quote.pm
> [@Basic/ExtraTests] rewriting author test xt/author/pod-syntax.t
> [DZ] writing Finance-Quote in Finance-Quote-1.21
> [MarkdownInRoot] overriding README.md in root
> [DZ] building archive with Archive::Tar::Wrapper
> [DZ] writing archive to Finance-Quote-1.21.tar.gz
> [DZ] built in Finance-Quote-1.21
> ```

Scratch that. I have all the tags in my repo. nm...

allen-ziegenfus:
I pushed all the tags to my fork in case that helps.

